PR #1, opened 2017-11-08
Default values

eddyStreamlabs:
Make sure we never have null or undefined input values when creating the sources.


computerquip-streamlabs:
Merged manually due to forced push onto staging.

PR #16, opened 2018-01-22
Fix rare crash on SSE capable computers (everyone)

Xaymar:
This crash happens every 1 in 100000 times in obs-stream-effects 64-bit, which has the same GS::VertexBuffer wrapper as node-obs::Display and Face Masks. The crash happens due to unlucky placement of the buffers, which need to be 16-byte aligned, otherwise we write into memory that we shouldn't have written to. This results in funky crashes down the road which make absolutely no sense at all with a mdmp or a simple stack trace, since the cause of the crash happened earlier.

Affects pretty much all areas, as the Display was actively writing into memory of other areas. This might explain some exotic behavior we have seen, such as crashes inside the Chromium runtime or electron crashes.


Xaymar:
Tagging @granolaboy for Face Masks, which is also affected by the same misalignment crash. May require some extensive rework due to the deeper use of the wrapper in face masks.

eddyStreamlabs:
Looks good to me.

PR #23, opened 2018-01-31
Drastically reduce CPU usage caused by Display code

Xaymar:
This unfortunately does not eliminate the costs of text rendering yet, but it does improve the other areas (outline, resize boxes, guidelines) which were also affected. Side effect is that the drawing code is now capable of full rotation, cropping should be easily added as well.

Todo:
- Don't rebuild vertex buffers for text, use shaders to change UVs.
- Make Text DPI aware as it's fixed 16px now, previous one varied with scene size - tiny on 4k monitors.
- Add toggle to turn off text and guidelines for Alex N.s changes.

### DIFF
--- a/src/gs-limits.h
+++ b/src/gs-limits.h
@@ -22,5 +22,5 @@
 
 namespace GS {
 	static const uint32_t MAXIMUM_VERTICES = 0xFFFFFFu;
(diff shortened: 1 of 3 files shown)


eddyStreamlabs:
Don't really see the gain on my machine but it seems to help a bit for other people. Don't see any side effects caused by these changes.

PR #25, opened 2018-02-05
Display cpu improvements fix

Xaymar:
Similar to the previous PR, except this one actually fixes the scaling. Text is a little bit larger now, 12x24 pixels per character. Change the 12 in `float_t pt = 12 * dp->m_previewToWorldScale.y;` if you need it to be smaller.


eddyStreamlabs:
Adjusted the text size, looks good to me now. Merging.

PR #30, opened 2018-02-09
Apply stream delay when starting the output stream

eddyStreamlabs:


eddyStreamlabs:
I need to do some changes on the frontend in order to put this PR in. The UX is currently weird as the frontend doesn't know that the stream will start X seconds later if the delay is active.

PR #48, opened 2018-05-04
Reduce the large amount of redundancy in the client

computerquip-streamlabs:
An example of how to reduce the large amount of redundancy in the client. This PR isn't meant to be pulled in but to show the difference in cleanliness, readability, and increased consistency in general of the code.

### DIFF
--- a/obs-studio-client/source/utility.hpp
+++ b/obs-studio-client/source/utility.hpp
@@ -17,6 +17,13 @@
 
 #pragma once
 
+#include <vector>
+#include <memory>
+#include <nan.h>
+
+#include "error.hpp"
+#include "controller.hpp"
+
 #ifdef __cplusplus
 #define INITIALIZER(f) \
         static void f(void); \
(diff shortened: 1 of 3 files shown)


obs-studio-client/source/utility.hpp
```diff
@@ -48,5 +55,41 @@
...
+        {
+                ErrorCode error = (ErrorCode)response[0].value_union.ui64;
+                if (error != ErrorCode::Ok) {
+                        Nan::ThrowError(Nan::New<v8::String>(response[0].value_str).ToLocalChecked());
+                        return false;
+                }
+        }
+
+        if (!response.size()) {
+                Nan::ThrowError(__FUNCTION_NAME__ ": Failed to make IPC call, verify IPC status.");
```
Xaymar: __FUNCTION_NAME__ is a macro and will always show "ValidateResponse" here. For a proper stack identification, use `dbghelp.dll`. Documentation for that is here: https://msdn.microsoft.com/en-us/library/windows/desktop/ms679291(v=vs.85).aspx.

obs-studio-client/source/utility.hpp
```diff
@@ -48,5 +55,41 @@
...
+                return false;
+        }
+
+        return true;
+}
+
+static FORCE_INLINE std::shared_ptr<ipc::client> GetConnection()
+{
+        auto conn = Controller::GetInstance().GetConnection();
+        if (!conn) Nan::ThrowError(__FUNCTION_NAME__ ": Failed to obtain IPC connection.");
```
Xaymar: Same as above.

computerquip-streamlabs:
If you wanted to use function_name, I should have kept it as a macro. I'm not implementing an entire platform-specific stack trace right now.

PR #61, opened 2018-07-04
Automatic formatting support

Xaymar:
This PR adds .clang-format support for obs-studio-client and obs-studio-server, thus allowing automatic formatting. On 9392518 currently formats 'global.cpp' in the following way:

```
Removed, see changed files.
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,100 @@
+# Basic Formatting
+BasedOnStyle: LLVM
+TabWidth: 4
+UseTab: ForIndentation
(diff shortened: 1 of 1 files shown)


Xaymar:
I have modified the formatting rules to be less strict, but there are still some problems with very long names that can't be fixed.

Another project I'm following recently added a CI test to see if the formatting of a commit/PR matches what clang-format gives back. If it didn't, it would fail the build. I'd like to implement something like that for this too if we decide to go with this formatting style.

PR #62, opened 2018-07-04
client: Upgrade to electron 2.0.4

Xaymar:
Updates the default iojs version to v2.0.4 from v1.8.3.


Xaymar:
Note that frontend currently still uses 2.0.3 from what I can see, which is one hotfix earlier and contains some rather critical bugs. We should also update frontend if it hasn't been done yet.

PR #67, opened 2018-08-07
Datalane ipc

eddyStreamlabs:


Xaymar:
It looks this is now working correctly. We can merge this back into staging now, if there are no further issues that come from lib-datalane or lib-streamlabs-ipc.

PR #68, opened 2018-08-08
Implement basic Tests

Xaymar:
This adds support for AppVeyor to run tests using a PowerShell framework that sends all information back to AppVeyor. You can track the live state of every test and even see any console output from it.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,13 +10,15 @@ environment:
   matrix:
     - RuntimeURL: https://atom.io/download/atom-shell
       RuntimeName: iojs
-      RuntimeVersion: v2.0.4
+      RuntimeVersion: v2.0.6
+      ElectronVersion: "2.0.6"
     - RuntimeURL: https://nodejs.org/download/release
       RuntimeName: node
       RuntimeVersion: v9.4.0
     - RuntimeURL: https://atom.io/download/atom-shell
       RuntimeName: iojs
-      RuntimeVersion: v3.0.0-beta.1
+      RuntimeVersion: v3.0.0-beta.4
+      ElectronVersion: "3.0.0-beta.4"
   NodeVersion: 9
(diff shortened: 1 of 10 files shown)


appveyor.yml
```diff
@@ -10,13 +10,15 @@ environment:
   matrix:
     - RuntimeURL: https://atom.io/download/atom-shell
       RuntimeName: iojs
-      RuntimeVersion: v2.0.4
+      RuntimeVersion: v2.0.6
+      ElectronVersion: "2.0.6"
```
computerquip-streamlabs: This should probably be removed and RuntimeVersion just be placed in quotations where necessary. Environment variables with quotation marks are an occupational hazard as well.

computerquip-streamlabs:
Where are all of the changes from yarn.lock coming from if we didn't modify package.json? 500+ lines seems excessive.

computerquip-streamlabs:
Can't say I'm a fan of the test cases having spaces in their script names. That's probably asking for trouble.

appveyor.yml
Xaymar: This is needed for the npm install -g electron@$Env:ElectronVersion command. There is no "v2.0.6" of electron, only "2.0.6".

Xaymar:
> Where are all of the changes from yarn.lock coming from if we didn't modify package.json? 500+ lines seems excessive.

Likely from the previous attempt to have electron as a dev dependency. I squashed to commit for it into another commit. I'll fix it if I figure out how.

> Can't say I'm a fan of the test cases having spaces in their file names. That's probably asking for trouble.

You can name the tests anything you want. Dealing with spaces in a name isn't any different from just using underscores, unless someone forgets to place proper double quotes in the test runner. I did not want to spend too much time adding name formatting and everything to the test runner, so I just allow everything that the file system also allows.

PR #84, opened 2018-08-30
Unicode support

eddyStreamlabs:


obs-studio-server/source/nodeobs_display.cpp
```diff
@@ -84,6 +84,19 @@ struct DestroyWindowMessageQuestion {
...
 
+static void HandleWin32ErrorMessage() {
+	DWORD dwErrorCode = GetLastError();
+	LPSTR lpErrorStr = nullptr;
+	DWORD dwErrorStrSize = 16;
+	DWORD dwErrorStrLen = FormatMessageA(
+		FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM,
+		NULL, dwErrorCode, LANG_USER_DEFAULT, lpErrorStr, dwErrorStrSize, NULL);
+	std::string exceptionMessage(
+		"Unexpected WinAPI error: " + std::string(lpErrorStr, dwErrorStrLen));
```
Xaymar-Streamlabs: A `sprintf` call with a dedicated `std::vector<char>` buffer might be better suited here. Adding two std::strings is somewhat expensive, though it probably doesn't matter as this is a crash routine.

PR #85, opened 2018-08-30
Attempt to fix the display issues

eddyStreamlabs:


obs-studio-server/source/nodeobs_display.cpp
```diff
@@ -127,7 +141,10 @@ void OBS::Display::SystemWorker() {
 					LocalFree(errorStr);
 					answer->success = false;
 				} else {
-					SetLayeredWindowAttributes(newWindow, 0, 100, LWA_ALPHA);
+					if (IsWindows8OrGreater()) {
+						SetLayeredWindowAttributes(newWindow, 0, 0, LWA_ALPHA);
```
Xaymar-Streamlabs: This sets the window alpha to 0 for Window 8 and newer. Is this intended? The correct value for a visible window would be 255, as shown here: https://docs.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-setlayeredwindowattributes

eddyStreamlabs: I don't think this is being correctly applied anyway, 0 or 255 doesn't make any difference. I think that the fact we are passing 0 as crKey resets the transparency style and makes it opaque. Gonna change it for the sack of it but it doesn't make any difference.

Xaymar-Streamlabs: For that to have any effect, only LAYERED would have to be specified. Combinations seem to make it do nothing.

PR #88, opened 2018-09-02
Add unicode support, mostly for win32

computerquip-streamlabs:
This adds unicode support by using wide character windows functionality.


obs-studio-client/source/controller.cpp
```diff
@@ -117,12 +152,28 @@ bool close_process(ProcessInfo pi) {
...
 
 std::string get_process_name(ProcessInfo pi) {
-	std::vector<char> name_buf(MAX_PATH + 2);
-	DWORD name_len = MAX_PATH + 1;
-	if (QueryFullProcessImageName((HANDLE)pi.handle, 0, name_buf.data(), &name_len)) {
-		return std::string(name_buf.data(), name_buf.data() + name_len);
+	/* 255 is a coincidental value actually given by GetVolumeInformation.
+	 * This value can be something other than 255 but 255 is the common
+	 * value. We're targetting desktops and the likelihood this turns into
+	 * an issue is almost none but should be noted at least. */
```
Xaymar-Streamlabs: At least for me, the maximum path allowed on C:\ is 32767, as returned by GetVolumeInformation()->lpMaximumComponentLength. Perhaps we should query GetVolumeInformation() here?

obs-studio-client/source/controller.cpp
```diff
@@ -24,13 +24,48 @@
...
 #pragma region Windows
 #ifdef _WIN32
 #include <windows.h>
 #include <direct.h>
 #include <wchar.h>
 
+/* Length is the number of characters, not bytes */
+namespace {
+
+std::string from_utf16_wide_to_utf8(const wchar_t *from, size_t length = -1)
```
Xaymar-Streamlabs: It looks like we might need these more often, perhaps put them in utility.hpp/cpp so they can be re-used?

obs-studio-client/source/controller.cpp
```diff
@@ -194,11 +245,20 @@ std::shared_ptr<ipc::client> Controller::host(std::string uri) {
...
-	std::string pid_path = "server.pid";
+	std::string pid_path;
+
 #ifdef _WIN32
-	std::vector<char> tmp(MAX_PATH + 2);
-	TCHAR tmp_len = GetTempPath(MAX_PATH + 1, tmp.data());
-	pid_path = std::string(tmp.data(), tmp.data() + tmp_len) + "\\" + pid_path;
+	std::basic_string<TCHAR> tmp;
+	tmp.resize(MAX_PATH + 1);
+	DWORD tmp_len = GetTempPath(MAX_PATH + 1, &tmp[0]);
```
Xaymar-Streamlabs: See above notice about GetVolumeInformation()->lpMaximumComponentLength

obs-studio-server/CMakeLists.txt
```diff
@@ -241,6 +241,8 @@ IF(WIN32)
 		VERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}.${PROJECT_VERSION_TWEAK}
 		SOVERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}.${PROJECT_VERSION_TWEAK}
 	)
+	target_compile_definitions(${PROJECT_NAME}
+		PRIVATE -D_UNICODE -DUNICODE)
```
Xaymar-Streamlabs: (Optional) -D is not needed here, you are already adding a definition (if CMake doesn't parse the -D, this will likely create a definition called D_UNICODE and DUNICODE).

obs-studio-client/source/controller.cpp
computerquip-streamlabs: I think you're confusing max path length with component length. Maximum path length on NTFS I believe is 255. Windows will generally not allow you to make a folder or file past somewhere along 236 characters though. See (this)[http://rextester.com/KJC3864] for an example output of GetVolumeInformation. 

If you want to make sure that's correct, that's fine, especially if someone installs onto a non-NTFS partitions (such as fat though I believe 255 there is still more than good enough).

obs-studio-client/source/controller.cpp
computerquip-streamlabs: That's not a bad idea. If we use them often, I should also cache those conversion objects as they can be horrifically slow to create.

obs-studio-client/source/controller.cpp
computerquip-streamlabs: GetTempPath will not return something higher than MAX_PATH + 1 (about 261 according to the documentation). If our path is longer than that, I'm not sure what the API will do exactly so my assumption is it will always be that length or lower. If that's not good enough, then we need to choose a different API.

obs-studio-client/source/controller.cpp
Xaymar-Streamlabs: I am talking about the total path length, which can be bigger than 255 on NTFS (starting with Windows 7, NTFS2). This is a policy that can be enabled for all drives but by default is only enabled for C:\ (as far as i know at least). For example, this is a valid path in PowerShell and other Long Paths capable programs:
```
\\?\D:\aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```
That's 742 characters for the entire path, far above the 255 character limit that started to stop existing with NTFS2.

Since `get_process_name` is not returning just the executable file name but actually the full binary path (for safety, probably should have renamed the function), this will break on something like this for example:

```
Works:
C:\Users\MichaelFabianDirks\AppData\Local\Programs\Streamlabs\Streamlabs OBS\app.asar.unpacked\node_modules\obs-studio-node\obs64.exe

Broken:
D:\Program Files (x86)\Automatisch Installiert\Meine Software\Freie Software\GPLv3\Media, Video & Audio\Streaming, Recording & Tools\Streamlabs\Streamlabs Open Broadcaster Software\Streamlabs Open Broadcaster Software (1)\app.asar.unpacked\node_modules\obs-studio-node\obs64.exe
```

Even though this case can be very unlikely, I think we should still handle that case and try with a 32kb buffer the 2nd time. Though the only apps I've seen actually fail with this (so far) are git, mercury/hg and chrome/firefox if the username is too long and the cache path exceeds 255 characters because of that.

obs-studio-client/source/controller.cpp
computerquip-streamlabs: Oh I see that name isn't name here, it's path. What a variable name...

obs-studio-client/source/controller.cpp
computerquip-streamlabs: Seems we can do this for this functionality: http://rextester.com/KJC3864
I have a working version in code as well...

PR #93, opened 2018-09-06
Refactor Controller.cpp

computerquip-streamlabs:
This compacts the code, removes the explicit use of "new" (which is actually cleaner and more optimal), and fixes a leak when compiled as non-Unicode as well.

### DIFF
--- a/obs-studio-client/source/controller.cpp
+++ b/obs-studio-client/source/controller.cpp
@@ -20,12 +20,13 @@
 #include "utility.hpp"
 #include <string>
 #include <sstream>
-#include <node.h>
-#include <iostream>
 #include <nan.h>
 #include <fstream>
 #include <locale>
 #include <codecvt>
+
+static std::string serverBinaryPath = "";
+static std::string serverWorkingPath = "";
 
 #pragma region Windows
 #ifdef _WIN32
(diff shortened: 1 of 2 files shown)


computerquip-streamlabs:
There's a bug in this as well where the returned string is +1 in size due to an extra null-terminator that isn't expected. *sigh*

computerquip-streamlabs:
I've added a myriad of different fixes and improvements to this PR as I noticed various areas that were just poorly written. This should be everything for this PR.

computerquip-streamlabs:
I've made more changes. This gets simplifies a lot of the Unicode handling, breaks some of the more complicated functions down into smaller re-usable functions or more readable functions, etc.

obs-studio-client/source/controller.cpp
```diff
@@ -237,80 +282,23 @@ std::shared_ptr<ipc::client> Controller::host(std::string uri) {
 	if (m_isServer)
 		return nullptr;
 
-	// Store info
-	std::string program = serverBinaryPath;
 	std::string commandLine = '"' + serverBinaryPath + '"' + " " + uri;
```
Xaymar-Streamlabs: This is slow and results in four separate re-allocations, consider also improving this with std::stringstream or a sprintf_s.

obs-studio-client/source/controller.cpp
```diff
@@ -193,37 +172,103 @@ bool is_process_alive(ProcessInfo pinfo) {
...
+
+	/* Note that this isn't a hack to use the same buffer,
+	 * it's explicitly meant to be used this way per MSDN. */
+	GetLongPathName(&tmp[0], &tmp[0], tmp_len);
+
+	return from_utf16_wide_to_utf8(tmp.data());
+#else
+	/* Non-unicode path won't give us above MAX_PATH ever. */
+	/* Shrink and remove the explicit null-terminator */
+	tmp.resize(MAX_PATH);
```
Xaymar-Streamlabs: We should probably opt for a wide-char to utf-8 conversion here anyway and explicitly call GetLongPathNameW.

obs-studio-client/source/controller.cpp
```diff
@@ -237,80 +282,23 @@ std::shared_ptr<ipc::client> Controller::host(std::string uri) {
...
-	std::basic_string<TCHAR> long_tmp;
-	long_tmp.resize(long_path_len);
-	GetLongPathName(&tmp[0], &long_tmp[0], long_path_len);
+	std::string workingDirectory =
+		serverWorkingPath.empty() ? get_working_directory() : serverWorkingPath;
 
-	pid_path.assign(from_utf16_wide_to_utf8(long_tmp.data(),long_path_len));
+	// Test for existing process.
+	std::string pid_path(get_temp_directory());
+	pid_path.append("server.pid");
```
Xaymar-Streamlabs: Is get_temp_directory() guaranteed to return a path with / (or \) at the end?

computerquip-streamlabs: Yeah, so contrary to MSDN's docs, where it gives the example "C:\TEMP", it would actually return "C:\TEMP\".

PR #97, opened 2018-09-11
Win7 aero disabled

eddyStreamlabs:


obs-studio-server/source/nodeobs_display.h
```diff
@@ -13,6 +13,9 @@
 #undef NOWINOFFSETS
 #endif
 #include <windows.h>
+#include <versionhelpers.h>
+#include <Dwmapi.h>
+#pragma comment (lib, "dwmapi.lib")
```
Xaymar-Streamlabs: Please do not use pragma to link additional libraries. Instead put them into the CMakeLists.txt.

PR #102, opened 2018-09-18
server: Update libobs to 22.0.3

Xaymar-Streamlabs:
libobs 22.0.3 includes our Facemask Plugin and various other fixes to obs-browser and obs code and plugins.


Xaymar-Streamlabs:
I am considering releasing two obs-studio versions in the future: One with facemask and one without.

PR #111, opened 2018-10-02
Replace stat with a unicode-portable version of stat

computerquip-streamlabs:


Xaymar-Streamlabs (approved):
Seems to compile fine on my end and actually works on the russian name profile.

PR #112, opened 2018-10-03
Fix invalidation of source id on release

computerquip-streamlabs:
If you call release, it doesn't guarantee that the object is destroyed, therefor the object shouldn't be invalidated (as it can still be correctly valid). For instance, creating a filter and adding it to a source will have two sources associated with the filter. Removing a reference from the filter will have the result of the filter still being valid and being destroyed only when source its attached to is destroyed.


Xaymar-Streamlabs:
Minor nitpick about the commit, replace `Fix` with `Remove`. Even though you technically fixed a bug, you actually only removed a line - this makes it easier to tell later on why and what this change does.

As this only affects the client (and hopefully not in a bad way), I think this is okay. The server uses signals to listed to everything relevant anyway.

Xaymar-Streamlabs (approved):

PR #113, opened 2018-10-04
Change from sentry.io to backtrace.io for obs-studio-node-server

computerquip-streamlabs:


Xaymar-Streamlabs (approved):
:+1: Sentry is neat with their total idea, but the lack of proper native support and not even storing a single mdmp is problematic for us.

PR #114, opened 2018-10-04
Refactor some of the bulk source creation functionality

computerquip-streamlabs:
An update for Typescript was needed as well here. The documentation dependencies were causing some issues so those were also removed here. Read each commit for more details.


computerquip-streamlabs:
I need to test this still so no rush.

PR #122, opened 2018-10-18
Source query hotkeys

RodrigoHolztrattner:
Automatic source hotkey detection.

### DIFF
--- a/obs-studio-client/source/isource.cpp
+++ b/obs-studio-client/source/isource.cpp
@@ -17,6 +17,7 @@
 
 #include "isource.hpp"
 #include <error.hpp>
+#include <functional>
 #include "controller.hpp"
 #include "obs-property.hpp"
 #include "properties.hpp"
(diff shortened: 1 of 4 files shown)


computerquip-streamlabs:
Commits with a title similar to "Small fixes" is a sure sign that commit needs to either be split apart into smaller individual commit or renamed to be more clear, perhaps with a body message indicating the fixes made.

The subject line of the commit message shouldn't be longer than 50 characters.

Commit subject lines should be in the form of an action. For instance instead of 

> Unnecessary line removed

it should be 
> Removed unnecessary line

We follow these guidelines loosely: https://chris.beams.io/posts/git-commit/

I do see commits that could be squashed together but I'd rather not squash everything together since there's commits here that do different things for different reasons. For instance, the random fixes and removals should be split into separate commits to help us bisect any issues that might arise.

RodrigoHolztrattner:
Thank you for the feedback, I will try to follow these guidelines on my next commits for sure!

RodrigoHolztrattner:
Rebase done with updated comments.

EddyGharbi (approved):

EddyGharbi:
Waiting on the current release to be out and then I will merge it.

PR #129, opened 2018-10-26
apply padding and keep aspect ratio 

summeroff:
a display padding value applies to X and calculated new padding value for Y to keep aspect ratio.


EddyGharbi:
Looks good 👍

PR #131, opened 2018-10-29
Fix the service variable from being used without being initialized first

RodrigoHolztrattner:
Just a small fix that updates the initialization order. The updateVideoStreamingEncoder() method that is called by the createVideoStreamingEncoder() method won't be using a null service variable anymore.


EddyGharbi:
Looks good to me, will merge once the release is out 👍

EddyGharbi (approved):

PR #132, opened 2018-10-30
Module

EddyGharbi:


obs-studio-server/source/osn-module.cpp
```diff
@@ -19,7 +19,7 @@
 #include "error.hpp"
 #include "shared.hpp"
 
-void osn::Module::Reigster(ipc::server&)
+void osn::Module::Reigster(ipc::server& srv)
```
computerquip-streamlabs: Misspelled name here though no biggy.

EddyGharbi: Yup missed that, just fixed it 👍

computerquip-streamlabs:
This looks alright. I'm not sure the best way to test this unfortunately though. Has any testing on it been done?

EddyGharbi:
Yeah it's a bit complicated to test it. I'm DMing you the plugin I tested against that is also whitelisted in the frontend.

obs-studio-server/source/nodeobs_api.cpp
```diff
@@ -427,14 +428,16 @@ void OBS_API::OBS_API_initAPI(
...
 	appdata_path        = args[0].value_str.c_str();
-	appdata_path += "/node-obs/";
 
 	/* libobs will use three methods of finding data files:
 	* 1. ${CWD}/data/libobs <- This doesn't work for us
 	* 2. ${OBS_DATA_PATH}/libobs <- This works but is inflexible
 	* 3. getenv(OBS_DATA_PATH) + /libobs <- Can be set anywhere
 	*    on the cli, in the frontend, or the backend. */
 	obs_add_data_path((g_moduleDirectory + "/libobs/data/libobs/").c_str());
+	slobs_plugin = appdata_path.substr(0, appdata_path.size() - 14);
```
computerquip-streamlabs: What is 14 here?

EddyGharbi: I could probably make it easier to read. The frontend is giving us the full appdata path of the app, I need to go one step up to only get the appdata path.

computerquip-streamlabs: I see. I can see how this is a bit problematic. I'm not sure how I feel on us relying on the frontend for this entirely. Ideally, we should probably use SHGetKnownFolder and FOLDERID_RoamingAppData. To clarify intent for now though, this should probably be something like `strlen("/slobs-client") + 1` instead of 14.

computerquip-streamlabs: Actually, re-reading this, I'm not sure it does what it intended. I'm going to look at this in a debugger.

computerquip-streamlabs: Oh, Apparently, the above is old and you changed this already to 13. It should be `strlen("/slobs-client")` instead of 13.

EddyGharbi: Yeah I can change that to clarify the intent. I would prefer to rely on what the frontend is passing us. If the frontend relies on a wrong appData, the backend should use what the frontend uses otherwise it will cause a lot of issues.

PR #133, opened 2018-10-30
Add checks for invalid DIrectX version

RodrigoHolztrattner:
An error will be issued in case the obs fails when creating the graphical adapter (or fail at loading its dll).
The client will throw an exception with the correct error message, this exception should be handled by the frontend.

### DIFF
--- a/obs-studio-server/source/nodeobs_api.cpp
+++ b/obs-studio-server/source/nodeobs_api.cpp
@@ -496,7 +496,14 @@ void OBS_API::OBS_API_initAPI(
 	obs_apply_private_data(private_settings);
 	obs_data_release(private_settings);
 
-	openAllModules();
+	std::string error;
+	if (!openAllModules(error)) {
+		rval.push_back(ipc::value((uint64_t)ErrorCode::Error));
+		rval.push_back(ipc::value(error));
+		AUTO_DEBUG;
+		return;
+	}
+
 	OBS_service::createStreamingOutput();
 	OBS_service::createRecordingOutput();
 
@@ -814,9 +821,12 @@ typedef std::basic_string<char, ci_char_traits> istring;
 
 /* This should be reusable outside of node-obs, especially
 * if we go a server/client route. */
-void OBS_API::openAllModules(void)
+bool OBS_API::openAllModules(std::string& err)
 {
-	OBS_service::resetVideoContext(NULL);
+	if (!OBS_service::resetVideoContext(NULL)) {
+		err = "D3D11 context error.";
+		return false;
+	}
 
 	std::string plugins_paths[] = {g_moduleDirectory + "/obs-plugins/64bit", g_moduleDirectory + "/obs-plugins"};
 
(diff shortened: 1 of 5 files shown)


obs-studio-server/source/nodeobs_api.cpp
```diff
@@ -814,9 +821,12 @@ typedef std::basic_string<char, ci_char_traits> istring;
 
 /* This should be reusable outside of node-obs, especially
 * if we go a server/client route. */
-void OBS_API::openAllModules(void)
+bool OBS_API::openAllModules(std::string& err)
 {
-	OBS_service::resetVideoContext(NULL);
+	if (!OBS_service::resetVideoContext(NULL)) {
```
computerquip-streamlabs: While this should suffice right now (and is better relatively speaking), there's various things that could go wrong here outside of a bad D3D11 context.

RodrigoHolztrattner: Do you think more checks would be enough or maybe a dedicated class to handle obs failures and format them to the frontend?

computerquip-streamlabs: See this as to what it could possibly return: https://github.com/obsproject/obs-studio/blob/047e586eafa43b139a4a605985592be75570fd61/libobs/obs.h#L351

Our API is a little unorthodox here unfortunately so it's a bit confusing, as far as design goes, where the check should properly go. 

If possible, we should just pass the value back to the frontend and let them handle it. This way they can emit a message or alert box to the user in whatever fashion they want. We shouldn't format it, we should let the frontend do that.

RodrigoHolztrattner:
I thought you meant adding checks on the other OBS functions, my bad.
I've modified the code to pass the numeric error directly (in case there is an error).

obs-studio-server/source/nodeobs_api.cpp
```diff
@@ -495,8 +495,15 @@ void OBS_API::OBS_API_initAPI(
 	obs_data_set_bool(private_settings, "BrowserHWAccel", true);
 	obs_apply_private_data(private_settings);
 	obs_data_release(private_settings);
+	
+	int videoError;
+	if (!openAllModules(videoError)) {
+		rval.push_back(ipc::value((uint64_t)ErrorCode::Error));
+		rval.push_back(ipc::value(videoError));
+		AUTO_DEBUG;
+		return;
```
EddyGharbi, line 507: We should return something to the frontend here, either a basic false value or an error code to tell the frontend something bad happened and let it handle it. What the frontend should probably do : popup a message to the user tell them what failed and what should be done.

EddyGharbi (changes requested):

obs-studio-server/source/nodeobs_api.cpp
RodrigoHolztrattner: Currently the client will receive the videoError (int) from the server . Since I'm returning the message as an error the 	ValidateResponse() method, on the client side, will throw the error (including the numeric videoError) to the JS, this throw can be handled by the frontend or I need to perform this using some callback?

EddyGharbi: `OBS_service_resetVideoContext` is used by the frontend in other places so it's good that it returns an error code now. But how it's written in your PR, OBS_API is not returning anything to the frontend. Both of of the functions need to return here. You can export an enum error code and use it for both.

EddyGharbi: NVM missed read. it seems like you are correctly returning the error code do the frontend 👍

EddyGharbi (approved):

PR #139, opened 2018-11-05
Revise build system and CI

computerquip-streamlabs:
Do not pull this, I'm still revising. I need to add in some functionality to allow local versions of libobs at the least, in a convenient manner.


js/module.d.ts
```diff
@@ -24,6 +24,14 @@ export declare const enum EDeinterlaceFieldOrder {
     Top = 0,
     Bottom = 1
 }
+export declare const enum EVideoCodes {
```
computerquip-streamlabs, line 27: This is from a previous PR that didn't compile typescript to definitions.

computerquip-streamlabs:
This is reading for testing, probably not ready to pull yet. In particular, need to make sure module loading still works and need to make sure that tagging properly pulls in signed binaries.

PR #141, opened 2018-11-06
Clean project files, fix some casting warnings and memory leaks

RodrigoHolztrattner:
Just a cleanup/anti-memory leak branch that is supposed to reduce the amount of warnings we currently have and add some obs memory release methods, reducing also the amount of memory leaks.


EddyGharbi (changes requested):
Could you split this PR into 2 separates ones? One for the warnings and the other one for the memory leaks?
I have a lot of comments and generally it's better to review if one branch/PR correspond to a one single purpose. Also it is easier / quicker to integrate and iterate if we do that.

RodrigoHolztrattner:
Sure

PR #142, opened 2018-11-06
Release all windows and obs loaded modules (dlls) on shutdown

RodrigoHolztrattner:
This PR should remove some problematic behavior when terminating the app and letting the GC release all dlls, it's better to no depend on it and do that manually.


computerquip-streamlabs:
This should be fine. By the way we can observe if modules actually get removed in a debugger while stepping through this. Some may not if they still have something with a reference to them. I'd like to look at that just to see what's actually happening here as a consequence.

computerquip-streamlabs (approved):

EddyGharbi:
Looks good, like the idea of just testing with the browser source for now as it seems to be one source of our problems on shutdown.

EddyGharbi (approved):

PR #143, opened 2018-11-07
Detach source signal when releasing it

RodrigoHolztrattner:
Adds a method to detach the source signal when releasing it, not leaking its memory in obs.


EddyGharbi (approved):
Looks good to me. Had a look at how is the behavior in OBS and it seems to be the same. It makes sense to me that when "destroy" is emitted we need to free all the data related to the source and the disconnect the signal handler.

PR #144, opened 2018-11-07
Add setting Force GPU as render device

pdamasceno:
- Exposed ForceGPUAsRenderDevice setting under Advanced Settings/Video
- Changed the export of NvOptimusEnablement to be based on the value of ForceGPUAsRenderDevice stored in the basic.ini file


obs-studio-server/source/main.cpp
```diff
@@ -53,6 +51,49 @@ extern "C" __declspec(dllexport) DWORD NvOptimusEnablement = 1;
...
+		return 1;
+	}
+
+	if (settingValue) {
+		// Disable high performance graphics rendering
+		return 0;
+	}
+
+	// Enable high performance graphics rendering
+	return 1;
```
EddyGharbi: Can we make it a bit easier to read? We could initialize the `settingValue` to true and only change the it if we successfully find the config file and find the setting in it. Then return the `settingValue` in all cases.

EddyGharbi (changes requested):

obs-studio-server/source/main.cpp
pdamasceno: I did the changes you asked and ended up changing the name of the setting to be able to return settingValue directly and don't make everything confusing. I believe it is better now.

obs-studio-server/source/nodeobs_configManager.cpp
```diff
@@ -172,6 +172,11 @@ void initBasicDefault(config_t* config)
 	config_set_default_string(config, "Video", "ColorSpace", "601");
 	config_set_default_string(config, "Video", "ColorRange", "Partial");
 
+	// Set default value for EnableGPUAsRenderDevice
+	if (!config_has_user_value(config, "Video", "ForceGPUAsRenderDevice")) {
+		config_set_bool(config, "Video", "ForceGPUAsRenderDevice", true);
```
EddyGharbi: Sorry I miss this during my first review but I would use `config_set_default_bool` here in any cases, this is what it is meant for.

EddyGharbi (changes requested):

obs-studio-server/source/nodeobs_configManager.cpp
pdamasceno: I did the change and tested everything. It is working successfully.

PR #148, opened 2018-11-08
Add release methods for some obs objects

RodrigoHolztrattner:
Reduces the total amount of obs memory leaks


EddyGharbi (approved):
Good job finding these leaks!

PR #149, opened 2018-11-08
Add casts and defines to reduce the amount of warnings

RodrigoHolztrattner:
Reduces the amount of warnings around the code (add a lot of casts and some defines)


EddyGharbi:
Looks good 👍

EddyGharbi (approved):

PR #159, opened 2018-11-16
Initialize some service pointer variables to nullptr

RodrigoHolztrattner:
These pointers, if left uninitialized, could lead into an undefined behavior.

### DIFF
--- a/obs-studio-server/source/nodeobs_service.cpp
+++ b/obs-studio-server/source/nodeobs_service.cpp
@@ -4,13 +4,13 @@
 #include "error.hpp"
 #include "shared.hpp"
 
-obs_output_t*  streamingOutput;
-obs_output_t*  recordingOutput;
-obs_encoder_t* audioStreamingEncoder;
-obs_encoder_t* audioRecordingEncoder;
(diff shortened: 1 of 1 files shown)


obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -4,13 +4,13 @@
...
 #include "shared.hpp"
 
-obs_output_t*  streamingOutput;
-obs_output_t*  recordingOutput;
-obs_encoder_t* audioStreamingEncoder;
-obs_encoder_t* audioRecordingEncoder;
-obs_encoder_t* videoStreamingEncoder;
-obs_encoder_t* videoRecordingEncoder;
-obs_service_t* service;
+obs_output_t*  streamingOutput			= nullptr;
```
computerquip: The spacing here could probably use some alignment.
EDIT: To clarify, when aligning things like this, use spaces only. Tabs are of variable size so may look different for different people.

RodrigoHolztrattner: Corrected!

PR #162, opened 2018-11-19
Update hotkey handling

RodrigoHolztrattner:
Implements a method that will expose all current loaded hotkeys for the obs.
Removes the object ID parameter from the ProcessHotkeyStatus().
Move these hotkey methods to the API class.


obs-studio-client/source/nodeobs_api.cpp
```diff
@@ -34,10 +34,12 @@ void api::OBS_API_initAPI(const v8::FunctionCallbackInfo<v8::Value>& args)
 			return;
 		}
 	}
-
+
```
EddyGharbi: Can you remove that extra line and space?

obs-studio-client/source/nodeobs_api.cpp
```diff
@@ -34,10 +34,12 @@ void api::OBS_API_initAPI(const v8::FunctionCallbackInfo<v8::Value>& args)
 			return;
 		}
 	}
-
+	
 	args.GetReturnValue().Set(v8::Number::New(args.GetIsolate(), response[1].value_union.i32));
 }
 
+#include "isource.hpp"
```
EddyGharbi: I would keep putting includes at the top of the files, easier to read.

obs-studio-server/source/nodeobs_api.cpp
```diff
@@ -622,6 +627,129 @@ void OBS_API::OBS_API_getPerformanceStatistics(
...
+    void*                          data,
+    const int64_t                  id,
+    const std::vector<ipc::value>& args,
+    std::vector<ipc::value>&       rval)
+{
+	obs_hotkey_id hotkeyId = args[0].value_union.ui64;
+	uint64_t      press    = args[1].value_union.i32;
+
+	// TODO: Check if the hotkey ID is valid
+	obs_hotkey_enable_callback_rerouting(true);
```
EddyGharbi: Do we need to set that to false at some point or obs does it when the callback is done with its work?

EddyGharbi (changes requested):

obs-studio-client/source/nodeobs_api.cpp
RodrigoHolztrattner: Done

obs-studio-client/source/nodeobs_api.cpp
RodrigoHolztrattner: Removed it (unnecessary include), I was probably doing some testing and forgot to put it at the include section.

obs-studio-server/source/nodeobs_api.cpp
RodrigoHolztrattner: Moved it to the obs initialization, we just need to enable it one time.

EddyGharbi (approved):

PR #163, opened 2018-11-20
Add missing error and return statements

RodrigoHolztrattner:
=> This PR is on hold until we have our new frontend proxy working.

Add some missing boolean checks when creating obs resources.
Add missing error statements that could lead into the main app ignoring a backend error.


EddyGharbi (approved):
Looks good

PR #168, opened 2018-11-28
Remove the delayed loading of some dlls

RodrigoHolztrattner:
We aren't using this anymore so there is no reason to maintain this on our code.
Also, suddenly loading a lot of dlls isn't healthy .


computerquip-streamlabs:
Replaced by another PR to resolve conflict.

PR #169, opened 2018-11-30
Add missing return validations and statements

RodrigoHolztrattner:
Add missing return statements.
Update the old return verifications with the ValidateResponse() method.
Update the ValidateResponse() to handle the InvalidReference case.


obs-studio-client/source/utility.hpp
```diff
@@ -77,8 +77,8 @@
 
 #define dstr(s) #s
 #define vstr(s) dstr(s)
-// FORCE_INLINE
-static  bool ValidateResponse(std::vector<ipc::value>& response)
+
+static FORCE_INLINE bool ValidateResponse(std::vector<ipc::value>& response)
```
computerquip-streamlabs: I don't know why we have a force_inline utility attribute. This likely brings worse performance. Instead, provide inline and let the compiler decide whether or not to inline it. Or better yet, specify nothing and since the function has static linkage in whatever TU includes it, the compiler will probably inline it anyways.

RodrigoHolztrattner: Removed it!

EddyGharbi (approved):

PR #175, opened 2018-12-08
Add sentry support

RodrigoHolztrattner:
Add support for Sentry logging instead using Crashpad with Backtrace.
Depends on the "Add custom callstack handling" PR.


computerquip-streamlabs:
Those crow dependencies make me nervous. There's one called "cygcurl" which is probably a cygwin dependency. I'm not sure how it pulls in dependencies but we should at the least include only what we need.

In addition, I think it's preferable to store binaries in an external archive then downloaded (similar to how obs has you download a dependency archive). Keeping binaries in the git repo itself adds unnecessary weight later on.

RodrigoHolztrattner:
> Those crow dependencies make me nervous. There's one called "cygcurl" which is probably a cygwin dependency. I'm not sure how it pulls in dependencies but we should at the least include only what we need.
> 
> In addition, I think it's preferable to store binaries in an external archive then downloaded (similar to how obs has you download a dependency archive). Keeping binaries in the git repo itself adds unnecessary weight later on.

I agree, my idea was adding them as sub-modules but I was having a hard time with the cmake files, I decided to do this way most because we still need to see if this will be a permanent feature, not sure how switching to Sentry will perform instead using Backtrace.
But I plan to modify this in the future if we found it a good addition and if we keep the Sentry support.

PR #183, opened 2018-12-14
Add validations to prevent calling strcmp with a null argument

RodrigoHolztrattner:
Make sure we won't ever call strcmp with a null argument when querying encoder types, doing this could lead into undefined behavior.
Add a try-catch block  that should prevent a crash when performing an obs call to an invalid encoder, this is supposed to avoid a startup crash.


EddyGharbi (approved):
I'm not sure the try catch will ever catch anything but otherwise looks good.

PR #196, opened 2019-01-02
Fix a crash source when performing logs with an invalid log file

RodrigoHolztrattner:
Fix a crash when attempting to use an invalid file handler, this happens when the log file creation fails and we try to write to it when logging.
The crash report is this: https://sentry.io/streamlabs-obs/obs-server/issues/693172838/?query=is%3Aunresolved

Basically the file creation check is wrong:
`if (!logfile)`
this would never hit since it's a ptr (even a failed new wouldn't set the ptr to nullptr), the correct way to perform it is using:
`if (!logfile->is_open())`


EddyGharbi (approved):
Looks good

PR #197, opened 2019-01-02
Fix directory path crash

RodrigoHolztrattner:
Fix a possible attempt to call the obs mkdir method with an invalid path argument.
Crash report: https://sentry.io/streamlabs-obs/obs-server/issues/696089066/?query=is%3Aunresolved

This also removes a memory leak on `os_dir_t* dir = path && path[0] ? os_opendir(path) : nullptr;`, here `dir` was unused.
(we had 2 lines equal that one but only one of them was actually leaking, the second was doing it right, releasing the `dir` after opening it. I removed both since we weren't using the variable on both cases anyway).

### DIFF
--- a/obs-studio-server/source/nodeobs_service.cpp
+++ b/obs-studio-server/source/nodeobs_service.cpp
@@ -1683,12 +1683,13 @@ void OBS_service::updateAdvancedRecordingOutput(void)
     bool noSpace = 
 		config_get_bool(ConfigManager::getInstance().getBasic(), "AdvOut", "RecFileNameWithoutSpace");
 
-	os_dir_t* dir = path && path[0] ? os_opendir(path) : nullptr;
-
-	os_closedir(dir);
+	string initialPath;
+	if (path != nullptr) {
+		initialPath = path;
+	}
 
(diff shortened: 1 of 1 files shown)


obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -1594,14 +1597,17 @@ void OBS_service::updateRecordingOutput(bool updateReplayBuffer)
...
+	}
 
     if(filenameFormat == NULL) {
         filenameFormat = "%CCYY-%MM-%DD %hh-%mm-%ss";
     } 
-    string strPath;
-    strPath += path;
 
+    string strPath;
+	strPath += initialPath;
```
EddyGharbi, line 1610: Could you remove this extra space?

EddyGharbi (changes requested):

obs-studio-server/source/nodeobs_service.cpp
RodrigoHolztrattner: Done

EddyGharbi (approved):

PR #202, opened 2019-01-07
[WIP] Add proper encoder ref increments/decrements

RodrigoHolztrattner:
Fix the usage of our Encoders by adding all the `add_encoder_ref` and `release_encoder` methods, until now they had problems (by releasing multiple times the same encoder) that could lead into crashes.

The idea is to simulate the RAII, if we request an encoder we increment the ref counting, when that encoder variable isn't used we decrement the ref counting, this way we will ensure it will only be deleted when it's not in use.


obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -682,20 +683,22 @@ const char* FindAudioEncoderFromCodec(const char* type)
...
 		return false;
 	}
 
-	if (audioEncoder != NULL && usingRecordingPreset)
-		obs_encoder_release(*audioEncoder);
-
+	obs_encoder_release(*audioEncoder);
 	*audioEncoder = obs_audio_encoder_create(id, "simple_audio", nullptr, 0, nullptr);
-	if (*audioEncoder == nullptr) {
+	if (*audioEncoder) {
```
summeroff: Is it expected to return false if audioEncoder recreated successfuly? 
It will not change anything for now as we not check return from this function in our code. But for future it looks inconsistent.

obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -711,12 +714,9 @@ bool OBS_service::createVideoStreamingEncoder()
...
     }
 
-    if(videoStreamingEncoder != NULL) {
-        obs_encoder_release(videoStreamingEncoder);
-    }
-
+	obs_encoder_release(videoStreamingEncoder);
 	videoStreamingEncoder = obs_video_encoder_create(encoder, "streaming_h264", nullptr, nullptr);
-	if (videoStreamingEncoder == nullptr) {
+	if (videoStreamingEncoder) {
```
summeroff: If we return on successful creation of encoder then `updateVideoStreamingEncoder` be called only for unsuccessful encoder creation.

obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -894,11 +894,9 @@ static void remove_reserved_file_characters(std::string& s)
 
 bool OBS_service::createVideoRecordingEncoder()
 {
-	if (videoRecordingEncoder != NULL) {
-		obs_encoder_release(videoRecordingEncoder);
-	}
+	obs_encoder_release(videoRecordingEncoder);
 	videoRecordingEncoder = obs_video_encoder_create("obs_x264", "simple_h264_recording", nullptr, nullptr);
-	if (videoRecordingEncoder == nullptr) {
+	if (videoRecordingEncoder) {
```
summeroff: Same as above. Why false for successful creation.

obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -1019,6 +1017,7 @@ bool OBS_service::startStreaming(void)
 		obs_data_t* settings     = obs_data_create();
 		obs_data_set_int(settings, "bitrate", audioBitrate);
 
+		obs_encoder_release(audioStreamingEncoder);
 		audioStreamingEncoder = obs_audio_encoder_create(id, "alt_audio_enc", nullptr, int(trackIndex) - 1, nullptr);
 		if (!audioStreamingEncoder)
```
summeroff: Looks like potential leak of `settings` as we return without releasing it. 
And few more places like this below.

obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -2097,50 +2104,76 @@ void OBS_service::UpdateRecordingSettings()
...
 {
+	obs_encoder_addref(videoStreamingEncoder);
 	return videoStreamingEncoder;
 }
 
 void OBS_service::setStreamingEncoder(obs_encoder_t* encoder)
 {
-	if (!videoStreamingEncoder)
-		obs_encoder_release(videoStreamingEncoder);
+	obs_encoder_release(videoStreamingEncoder);
```
summeroff, line 1977: In case when `videoStreamingEncoder` is the same as `encoder`. `obs_encoder_release` can destroy object and `obs_encoder_addref` be applied on destroyed object. 
Currently it can be non issue as no one yet call `setStreamingEncoder`  to set same encoder object. Or object still have few references and not destroyed right after `obs_encoder_release`. But when other code change this can become an issue. 
Or if someone call `setStreamingEncoder( getStreamingEncoder() )` it just increase ref count instead of doing nothing.

obs-studio-server/source/nodeobs_service.cpp
RodrigoHolztrattner: That's true, find&replace didn't added the `!` statements, I will fix this tomorrow, thanks

obs-studio-server/source/nodeobs_service.cpp
RodrigoHolztrattner: Added the data release, we probably would crash if it returns false but it's a good thing to have if in the future we change the behavior.

obs-studio-server/source/nodeobs_service.cpp
RodrigoHolztrattner: I will add some checks to prevent it from happening, not sure if an assert would be better to proper indicate invalid code structure when on dev environment.

RodrigoHolztrattner: A case like `setStreamingEncoder(getStreamingEncoder() )` should never happen since the user is responsible to release any pointer that he got by calling `getxxxxxxEncoder()`, a fix for this would be using RAII.

RodrigoHolztrattner:
This isn't necessary anymore since another PR will fix these issues.

PR #203, opened 2019-01-08
fix: issues found with analyzer

summeroff:
for problematic newserviceTypeValue changed from char * to std::string . Now it will be more like other var newServiceValue from same function with similar usage .


EddyGharbi (approved):
Looks good to me 👍

PR #206, opened 2019-01-09
Make osn-tests run in osn CI

pdamasceno:
- Integrated obs-studio-node-tests to the obs-studio-node branch. It is in tests/osn-tests
- Renamed obs-studio-node-tests to osn-tests
- Removed iojs v3.0.8 build from appveyor.yml as it was not being used
- Changed node version to be compatible with iojs v2.0.8
- Added the compile and run of osn-tests in the test script of appveyor.yml


appveyor.yml
```diff
@@ -11,12 +11,8 @@ environment:
     - RuntimeURL: https://atom.io/download/atom-shell
       RuntimeName: iojs
       RuntimeVersion: v2.0.8
+      NodeVersion: 8
```
EddyGharbi, line 14: Why changing the node version? Also we should maybe use the LTS version.

EddyGharbi (changes requested):

appveyor.yml
pdamasceno: The iojs version v2.0.8 requires the node version 8 to run. Changed the node version to prevent any incompatibilities. Previously we were using version 9.

EddyGharbi: I see, then I'm confused on how it was working before.

pdamasceno: Maybe we were lucky to not have encountered any critical incompatibilities with the version 9. Looked into the LTS version and node version 8 is in maintenance LTS. It will last up to December this year
https://github.com/nodejs/Release

EddyGharbi (approved):

PR #208, opened 2019-01-10
[WIP] Fix many obs object leaks by adding their release methods

RodrigoHolztrattner:
Fix a lot of leaked obs objects. Reducing the amount of leaks and preventing possible crashes caused by destroying these same objects after obs was shutdown.


obs-studio-server/source/osn-volmeter.cpp
```diff
@@ -35,9 +35,10 @@ osn::VolMeter::VolMeter(obs_fader_type type)
 		throw std::exception();
 }
 
-osn::VolMeter::~VolMeter()
```
RodrigoHolztrattner, line 38: Fixes this shutdown crash https://sentry.io/streamlabs-obs/obs-server/issues/695709152/?query=is%3Aunresolved

PR #212, opened 2019-01-15
Remove `using namespace` statements from files

RodrigoHolztrattner:
Remove many statements like this that could conflict with other libraries and was a problem when updating to C++17.


EddyGharbi (approved):
Looks good to me, please resolve the merge conflicts so we can merge.

RodrigoHolztrattner:
Done!

PR #213, opened 2019-01-15
Fix fader volmeter module generation

RodrigoHolztrattner:
This PR is an attempt to fix how the V8 library is generating the modules for these 2 classes, it seems that it is not being able to proper generate the code and it could cause some crashes when creating/deleting an object from them. It's probably related to our old V8 library usage.

Update1: The current implementation will make the unity tests work.
Update2: This PR is frozen until we decide if it's a good approach to resolve this error.


pdamasceno (approved):
We already discussed this extensively. Everything is ok now that we are not using a unique pointer in these cases.

EddyGharbi:
What's the status on this?

RodrigoHolztrattner:
Fixed the conflicts. This will allow our tests to pass and I added a task on Asana to review it in the future https://app.asana.com/0/734357654754972/1108852507583058/f

EddyGharbi:
Looks good 👍

PR #219, opened 2019-01-18
Improve sentry report with custom annotations

RodrigoHolztrattner:
Add many features aimed to enrich our reports:

- Custom callstack rewind (avoid damaged/corrupted mini-dumps)
- CPU/RAM usage (situations where a crash could be explained by extreme usage of pc resources)
- OBS Log - last 150 lines from the log file (maximum supported by Sentry atm)
- OBS Total object leaks
- OBS Status (initialized/de-initialized, good for catching bugs on startup/shutdown)
- Time elapsed (match value from backtrace)
- Process list (still on early dev stage, for now it's just a list of active processes but in the future it will give us the complete SLOBS tree, useful for finding zombie processes and helping determining crash sources)

This PR basically moves all crashpad functionality to a proper class (`util-crashmanager`), the crashpad is instanced when starting SLOBS and when a crash occurs, this last instance will substitute the first one and it will contain this new information. This is a well established operation for crashpad since is the only way to add annotations at runtime (if the second instance fails or SLOBS crashes during its initialization, the first one will generate a report without including the new annotations).

Crashpad will still be disabled for debug builds on this PR.


obs-studio-server/source/util-crashmanager.cpp
```diff
@@ -0,0 +1,745 @@
...
+	physMemUsed     = long long(-1);
+	physMemUsedByMe = size_t(-1);
+	totalCPUUsed    = double(-1.0);
+
+#endif
+}
+
+void GetUserInfo(std::string& computerName)
+{
+#define INFO_BUFFER_SIZE 32767
```
summeroff: GetComputerName requires buffer size to be MAX_COMPUTERNAME_LENGTH +1 which is smaller than 32767.

obs-studio-server/source/nodeobs_api.cpp
```diff
@@ -328,6 +329,9 @@ static void                                    node_obs_log(int log_level, const
 			// File Log
 			*logStream << newmsg << std::flush;
 
+            // Internal Log
+			logMessages.push_back(newmsg);
```
summeroff: We need only last 150 messages so we can limit size of logMessages to not waste memory keeping all messages. 
Something like ```if(logMessages.size() > 150) logMessages.pop_front();```

And `std::list` or `std::queue` probably be better than `std::vector` with such push pop access. 

Also it is a bad practice to hardcode 150 here so change it to be configurable. Maybe function like `setOBSInternalLogSize(int logsize)`

RodrigoHolztrattner: The reason it is 150 is that Sentry doesn't handle well more then 150 lines when processing annotations, it messes its JSON formation and hides most on the information cell. This log is supposed to be used for other purposes then reports so it's better to not limit it (at least right now).
I'm still in the process to find a way to increase the Sentry limit (if possible to support the entire log).

obs-studio-server/source/util-crashmanager.cpp
RodrigoHolztrattner: Changed it, thanks for catching. Bad copy & paste code.

obs-studio-server/source/nodeobs_api.cpp
summeroff: I still think there should be some limit if we keep log in ram. Users run program for days some time and without limit this log will only grow till no free memory left. 
Also log still saved to filestream in `*logStream << newmsg << std::flush;` and can be read from there if needed more for some other purpose.

RodrigoHolztrattner: I've updated how the log stuff is done, I'm also limiting the total amount of messages stored by 150 but the error and warnings are still stored in a vector (since their number won't be huge).

The problem about using the log file is that some users aren't being able to save the file on their machines, on these cases it would be impossible to use the file itself (and I'm planning into sending the log even if the user was usable to generate the file), but for now I will limit the amount of data stored and in revise this in the future, when/if necessary.

EddyGharbi (approved):

PR #220, opened 2019-01-18
Add remaining osn-global functions test cases

pdamasceno:
* Add test cases for remaining osn-global functions
* Change npm to yarn being used to start tests in order to maintain
package manager consistency
* Add general.ts file that has helper functions used by all tests
* Removed word "should" from the test cases text
* Reworked all tests to be more in line with how the functions are being used in the frontend
* Added missing tsconfig.json to compile typescript files properly

### DIFF
--- a/tests/osn-tests/src/test_nodeobs_api.ts
+++ b/tests/osn-tests/src/test_nodeobs_api.ts
@@ -33,15 +33,13 @@ describe('nodebs_api', () => {
         }
     });
 
-    after(function(done) {
-        this.timeout(5000);
+    after(function() {
         obs.shutdown();
         obs = null;
-        setTimeout(done, 3000);
     });
 
     context('# OBS_API_getPerformanceStatistics', () => {
-        it('should fill stats object', () => {
+        it('Fill stats object', () => {
             let stats: IPerformanceState;
 
             try {
@@ -59,7 +57,7 @@ describe('nodebs_api', () => {
     });
 
     context('# OBS_API_QueryHotkeys', () => {
-        it('should get all hotkeys', () => {
+        it('Get all hotkeys', () => {
             try {
                 obsHotkeys = osn.NodeObs.OBS_API_QueryHotkeys();
             } catch(e) {
@@ -69,7 +67,7 @@ describe('nodebs_api', () => {
     });
 
     context('# OBS_API_ProcessHotkeyStatus', () => {
-        it('should process all hot keys gotten previously', () => {
+        it('Process all hot keys gotten previously', () => {
             let hotkeyId: any;
             let isKeyDown: boolean;
 
@@ -82,7 +80,7 @@ describe('nodebs_api', () => {
             }
         });
 
-        it('should throw error if hot key id does not exist', () => {
+        it('FAIL TEST: Try to process hot key id that does not exist', () => {
             let isKeyDown: boolean;
 
             expect(function() {
--- a/tests/osn-tests/src/test_nodeobs_api.ts
+++ b/tests/osn-tests/src/test_nodeobs_api.ts
@@ -1,7 +1,8 @@
 import 'mocha'
 import { expect } from 'chai'
 import * as osn from 'obs-studio-node';
-import { OBSProcessHandler } from '../util/obs_process_handler'
+import { OBSProcessHandler } from '../util/obs_process_handler';
+import { getCppErrorMsg } from '../util/general';
 
 interface IPerformanceState {
     CPU: number;
@@ -46,7 +47,7 @@ describe('nodebs_api', () => {
             try {
                 stats = osn.NodeObs.OBS_API_getPerformanceStatistics();
             } catch(e) {
-                throw new Error("OBS_API_getPerformanceStatistics threw an expection");
+                throw new Error(getCppErrorMsg(e));
             }
             
             expect(stats.CPU).to.not.equal(undefined);
@@ -62,7 +63,7 @@ describe('nodebs_api', () => {
             try {
                 obsHotkeys = osn.NodeObs.OBS_API_QueryHotkeys();
             } catch(e) {
-                throw new Error("OBS_API_QueryHotkeys threw an expection");
+                throw new Error(getCppErrorMsg(e));
             }
         });
     });
@@ -76,7 +77,7 @@ describe('nodebs_api', () => {
                 try {
                     osn.NodeObs.OBS_API_ProcessHotkeyStatus(hotkeyId, isKeyDown);
                 } catch(e) {
-                    throw new Error("hotkeyId " + hotkeyId + "does not exist");
+                    throw new Error(getCppErrorMsg(e));
                 }
             }
         });
--- a/tests/osn-tests/src/test_osn_filter.ts
+++ b/tests/osn-tests/src/test_osn_filter.ts
@@ -1,8 +1,9 @@
 import 'mocha'
 import { expect } from 'chai'
-import { OBSProcessHandler } from '../util/obs_process_handler'
 import * as osn from 'obs-studio-node';
 import { IFilter } from 'obs-studio-node';
+import { OBSProcessHandler } from '../util/obs_process_handler';
+import { getCppErrorMsg } from '../util/general';
 
 describe('osn-filter', () => {
     let obs: OBSProcessHandler;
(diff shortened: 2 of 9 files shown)


tests/osn-tests/src/test_osn_filter.ts
```diff
@@ -24,17 +25,31 @@ describe('osn-filter', () => {
...
+        it('should get all filter types', () => {
+            let filterTypes: string[];
+
+            try {
+                filterTypes = osn.FilterFactory.types();
+            } catch(e) {
+                throw new Error(getCppErrorMsg(e));
+            }
+
+            expect(filterTypes.length).to.not.equal(0);
```
EddyGharbi: I would test more things here. Only making sure the array is not empty doesn't tell us if the function is correctly working or not.

tests/osn-tests/src/test_nodeobs_api.ts
```diff
@@ -46,7 +47,7 @@ describe('nodebs_api', () => {
             try {
                 stats = osn.NodeObs.OBS_API_getPerformanceStatistics();
             } catch(e) {
-                throw new Error("OBS_API_getPerformanceStatistics threw an expection");
+                throw new Error(getCppErrorMsg(e));
```
EddyGharbi, line 48: I would put these kind of changes in a separate PR to facilitate the work for the reviewer.

tests/osn-tests/src/test_osn_global.ts
```diff
@@ -24,23 +25,29 @@ describe('osn_global', () => {
     });
 
     context('# SetOutputSource', () => {
-        it('should set source to output channel', () => {
-            let source: IInput;
+        let source: IInput;
 
+        it('should set source to output channel', () => {
```
EddyGharbi: general feedback: I would change `should set source to output channel` to `Set source to output channel`, clearer and we repeat less across the tests.

tests/osn-tests/src/test_osn_global.ts
```diff
@@ -50,10 +57,92 @@ describe('osn_global', () => {
...
             expect(source.name).to.equal('test_osn_global_source');
         });
+
+        it('should fail if trying to get from empty output channel', () => {
+            let source: ISource;
+
+            try {
+                source = osn.Global.getOutputSource(5);
+            } catch(e) {
+                throw new Error(getCppErrorMsg(e));
```
EddyGharbi: We have to ask our self if the way it fails is expected by the frontend. Throwing an error when something wrong happened is good but if nobody is handling that error then we just end up in the same bad state.

tests/osn-tests/src/test_osn_global.ts
```diff
@@ -50,10 +57,92 @@ describe('osn_global', () => {
...
+        });
+    });
+
+    context('# LaggedFrames', () => {
+        it('should get lagged frames value', () => {
+            let laggedFrames: number;
+
+            expect(function() {
+                laggedFrames = osn.Global.laggedFrames;
+            }).to.not.throw();
```
EddyGharbi: Good to test that but I'm not sure we are testing the correct behavior of the function here.

tests/osn-tests/src/test_osn_global.ts
```diff
@@ -50,10 +57,92 @@ describe('osn_global', () => {
...
+
+    context('# GetLocale', () => {
+        it('shoud get locale', () => {
+            let locale: string;
+
+            expect(function() {
+                locale = osn.Global.locale;
+            }).to.not.throw();
+
+            expect(locale).to.equal('pt-BR');
```
EddyGharbi, line 127: I believe this is true on your machine, not on others. Let's set English so that it passes for everyone + Appveyor.

EddyGharbi (changes requested):

tests/osn-tests/src/test_nodeobs_api.ts
pdamasceno: I'll keep that in mind for future PRs, this change was made in the commit with a whole lot of stuff so I can't remove it right now without messing up the whole branch.

tests/osn-tests/src/test_osn_global.ts
pdamasceno: Removed the word 'should' from the text of all test cases and I'll not used this word anymore in the future

tests/osn-tests/src/test_osn_global.ts
pdamasceno: The purpose of these tests is to guarantee that our functions are working correctly in the most simple test case possible. Here I'm throwing the error that comes from the backend to make the test case fail in case something went wrong on our side. I'm not factoring how the frontend is expected to behave in case of success or fails because it would increase the scope of the unit testing beyond what is being proposed by these tests as well as the time to have full test coverage.

tests/osn-tests/src/test_osn_global.ts
pdamasceno: Changed it. Now I initialize the value as undefined and after getting the value I check if it is still undefined.

tests/osn-tests/src/test_osn_global.ts
pdamasceno: The OBS process is indeed started with en-US as locale. Here I'm just testing if the function GetLocale is returning the value as expected. In the previous test case I used SetLocale to change it to 'pt-BR'.

tests/osn-tests/src/test_osn_filter.ts
pdamasceno: I tried to add and expect to compare the content of the array with the content of another array I initialized but the difference in the filters available on each machine makes this impossible to work. For example, in my machine I have ndi_filter, ndi_audiofilter and permultiplied_alpha_filter and in the machine that the AppVeyor runs these 3 filters are missing. Any other ideas as to what I should test here?

EddyGharbi: What we should do is make sure the array we fetch contains a specific set of values.

tests/osn-tests/src/test_nodeobs_api.ts
EddyGharbi: Sounds good for this one 👍

tests/osn-tests/src/test_osn_global.ts
EddyGharbi: Sounds good, thanks

tests/osn-tests/src/test_osn_global.ts
EddyGharbi: My point is that should we except that behavior in case of error? We should have a proper way to handle errors in the frontend and our backend should follow it. Then in our tests, we should test that our backend behaves correctly when an error occurred.

tests/osn-tests/src/test_osn_global.ts
EddyGharbi: If some tests are depending from others we should make sure it is done in an explicit way. Otherwise it's gonna be very hard for others to maintain these tests and add others.

tests/osn-tests/src/test_osn_filter.ts
pdamasceno: Figured out what the basic OBS filter types are and now I verify if the array has them

tests/osn-tests/src/test_osn_global.ts
pdamasceno: Reworked all tests to be more in line with how the functions are used in the frontend. I'll have this in mind for all future tests

tests/osn-tests/src/test_osn_global.ts
pdamasceno: Changed the SetLocale and GetLocale to be in the same test case. Now it is more clear what is happening and I'll pair Set and Get functions together from now on. It makes more sense to have them together.

tests/osn-tests/src/test_osn_global.ts
```diff
@@ -16,44 +18,115 @@ describe('osn_global', () => {
...
     });
 
+    // Shutdown OBS process
     after(function(done) {
-        this.timeout(3000);
+        this.timeout(5000);
         obs.shutdown();
         obs = null;
-        setTimeout(done, 2000);
+        setTimeout(done, 3000);
```
EddyGharbi: I'm really really not a fan of all of these timeouts. I don't get why we need them.

pdamasceno: If there is no timeout there, when the next test file starts running there will be two OBS process in the system. The calls to obs function can use the OBS process being shutdown and everything breaks if this happens. The timeout is to give time to the OBS process to shutdown before starting a new one.

EddyGharbi: We should be listening to an event here, we should force that to be a synchronous operation.

pdamasceno: Created a promise with IPC.disconnect() and used await to wait it complete since disconnect is async. Removed all the setTimeouts and the tests run fine with the OBS process being created and closed properly. That is enough for this or I'm doing something wrong here?

pdamasceno: Remove the above. There is a separate PR that makes the disconnect function wait for the server to shutdown properly. Tested those changes with this branch, with the timeouts on shutdown removed, and the OBS process starts and closes as intended. Waiting to see if there is no change requests to PR #230 to merge it with this one.

EddyGharbi (approved):
Looks good to me 👍

PR #221, opened 2019-01-22
Add global timeout to mocha initialization to prevent timeout failure of hooks and test cases

pdamasceno:
The default value for mocha timeout is 2 seconds. If any hook or test case exceeds that time it will fail. Changed it to 10 minutes.

### DIFF
--- a/tests/osn-tests/package.json
+++ b/tests/osn-tests/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "description": "OBS Studio Node Unit Testing",
   "scripts": {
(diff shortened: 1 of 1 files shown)


EddyGharbi (commented):
There is a no timeout flag, why couldn't we use it?

pdamasceno:
Didn't tried with it. The timeout flag worked fine so I didn't try any other. I can change it to the no timeouts flag and see the if results are the same

EddyGharbi:
Can you try it? It should be a matter of minutes. This could have been done earlier.

EddyGharbi (approved):

PR #222, opened 2019-01-22
Update the sentry auth token for backend

RodrigoHolztrattner:
Change the current Sentry token to one who only have `release` permissions, this means that it will allows posting but not reading or retrieving any other info.


pdamasceno (approved):
Everything is ok!

EddyGharbi (approved):

PR #224, opened 2019-01-22
Add fader & volmeter shutdown releases

RodrigoHolztrattner:
Release these objects when performing the shutdown sequence, we still need to make the frontend release them whenever necessary (scene change for example) since no call to their respective `delete` method is being made.

This PR will also remove unused method on the client side that were used in the past to release them, now this is done only on the server side to avoid any possible crash related to invalid memory deletion.

This PR has a collateral addition to the `utility:manager` class, it will add the methods `for_each` and `clear`.


EddyGharbi (approved):
Looks good to me 👍

PR #230, opened 2019-01-26
Make Controller::disconnect() wait for server to close properly

pdamasceno:


obs-studio-client/source/controller.cpp
```diff
@@ -316,6 +316,12 @@ void Controller::disconnect()
 {
 	if (m_isServer) {
 		m_connection->call_synchronous_helper("System", "Shutdown", {});
+
+		// Waiting for server to close
+		while (is_process_alive(procId)) {
```
EddyGharbi, line 321: We cannot do that here. This is exposing to a potential infinite loop that would freeze the app entirely.

EddyGharbi:
Let's put the PR on hold while we figure out if we actually needs it.

pdamasceno:
This is not needed. Found out the reason of the lost IPC connection when starting a OBS process. We can cancel this PR.

EddyGharbi:
Closing it 👍

PR #232, opened 2019-01-29
add tests for osn transition

summeroff:


tests/osn-tests/src/test_osn_transition.ts
```diff
@@ -0,0 +1,150 @@
...
+                expect(transition).to.not.equal(undefined);
+                expect(transition.id).to.equal(transitionType);
+                expect(transition.name).to.equal(transitionName);
+                transition.release();
+            }
+        });
+
+    });
+
+    context('# createPrivate', () => {
```
pdamasceno: The frontend don't use CreatePrivate. I added this function to the list of functions not being used. I think it is best to remove this test case until we decide what to do with this function in the backend

tests/osn-tests/src/test_osn_transition.ts
```diff
@@ -0,0 +1,150 @@
...
+                expect(transition.id).to.equal(transitionType);
+                expect(transition.name).to.equal(transitionName);
+                transition.release();
+            }
+        });
+
+       
+    });
+
+    context('# set', () => {
```
pdamasceno: Add getActiveSource and clear to the description of context here to inform that those functions are also being tested here

pdamasceno (changes requested):
Minor changes that I proposed. Everything is great with the test cases of osn-transition.

tests/osn-tests/src/test_osn_transition.ts
```diff
@@ -0,0 +1,150 @@
...
+
+            // Create each transition type availabe passing settings parameter
+            for (transitionType of transitionTypes)
+            {
+                transition = osn.TransitionFactory.create(transitionType, transitionName, settings);
+
+                // Checking if transition was created correctly
+                expect(transition).to.not.equal(undefined);
+                expect(transition.id).to.equal(transitionType);
+                expect(transition.name).to.equal(transitionName);
```
pdamasceno, line 72: Verify if the setting you created the transition with is present in the list of settings. See how it is done in osn-input test case.

pdamasceno (approved):

PR #233, opened 2019-01-29
fix: out of array on big resolutions

summeroff:
User had resolution 4500x3000 and do/while loop goes till idx take value outside of vals. 

Add generated values with step of 0.5 if vals is not enough.

### DIFF
--- a/obs-studio-server/source/nodeobs_service.cpp
+++ b/obs-studio-server/source/nodeobs_service.cpp
@@ -626,8 +626,15 @@ int OBS_service::resetVideoContext(bool reload)
 		if (ovi.base_width > 1280 && ovi.base_height > 720) {
 			int idx = 0;
 			do {
-				ovi.output_width  = uint32_t(double(ovi.base_width) / vals[idx]);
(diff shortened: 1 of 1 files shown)


obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -626,8 +626,15 @@ int OBS_service::resetVideoContext(bool reload)
...
 			int idx = 0;
 			do {
-				ovi.output_width  = uint32_t(double(ovi.base_width) / vals[idx]);
-				ovi.output_height = uint32_t(double(ovi.base_height) / vals[idx]);
+				double use_val = 1.0;
+				if(idx < numVals)
+				{
+					use_val = vals[idx];
+				} else {
+					use_val = 3.0 + double(numVals - idx + 1)/2.0;
```
RodrigoHolztrattner: I would use `vals[numVals-1]` instead `3.0` just in case we decide to change the values someday.

summeroff: agreed.

EddyGharbi (approved):
Looks good to me 👍

PR #236, opened 2019-01-29
Release the display text texture

RodrigoHolztrattner:
Release it on the display destructor

### DIFF
--- a/obs-studio-server/source/nodeobs_display.cpp
+++ b/obs-studio-server/source/nodeobs_display.cpp
@@ -363,7 +363,7 @@ OBS::Display::~Display()
 		delete m_textVertices;
 
     if (m_textTexture) {
-		gs_texture_destroy(m_textTexture);
+        gs_texture_destroy(m_textTexture);
     }
 
 	m_boxLine = nullptr;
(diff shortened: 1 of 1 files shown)


obs-studio-server/source/nodeobs_display.cpp
```diff
@@ -361,6 +361,11 @@ OBS::Display::~Display()
 	obs_enter_graphics();
 	if (m_textVertices)
 		delete m_textVertices;
+
+    if (m_textTexture) {
+		gs_texture_destroy(m_textTexture);
+    }
+
```
EddyGharbi, line 368: Please correct indentation here.

EddyGharbi (changes requested):

obs-studio-server/source/nodeobs_display.cpp
RodrigoHolztrattner: Done!

EddyGharbi: I still see the same issue.

RodrigoHolztrattner: I think now it's fixed

EddyGharbi: Looks good ty

PR #238, opened 2019-01-30
Add a mutex to prevent races when logging

RodrigoHolztrattner:
Any thread can make a call to `blog` and races should be handled by using a mutex on our logging method.


EddyGharbi (approved):
Looks good to me 👍 Good job finding this.

EddyGharbi:
Tests are not passing here, could you have a look?

RodrigoHolztrattner:
It was missing the last test changes, merged staging and now they are passing

PR #242, opened 2019-02-04
Handle Dx11 internal string exception

RodrigoHolztrattner:
A few crashes are happening because the user had an invalid (or none) version of the DirectX compiler (see https://sentry.io/streamlabs-obs/obs-server/issues/870076998/?query=is%3Aunresolved) this wasn't filtered until now.

This PR should catch an exception and transform it in an error code that will be handled by the frontend.

Obs: The report above throws this exception below:
```
	throw "Could not find any D3DCompiler libraries. Make sure you've "
		"installed the <a href=\"https://obsproject.com/go/dxwebsetup\">"
		"DirectX components</a> that OBS Studio requires.";
```

### DIFF
--- a/obs-studio-server/source/nodeobs_api.cpp
+++ b/obs-studio-server/source/nodeobs_api.cpp
@@ -518,7 +518,6 @@ void OBS_API::OBS_API_initAPI(
 	}
 
 	OBS_service::createService();
-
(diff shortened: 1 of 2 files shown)


summeroff (approved):

EddyGharbi (approved):
Looks good to me 👍

PR #243, opened 2019-02-05
Log if the user got an invalid path on his plugins path

RodrigoHolztrattner:
https://app.asana.com/0/734357654754972/1107683930734842/f

This will help us to identify plugin path problems by outputting any invalid path to the log file.


EddyGharbi (approved):
Looks good to me 👍

PR #246, opened 2019-02-06
Use global video context instead a saved one

RodrigoHolztrattner:
Asana ticket: https://app.asana.com/0/734357654754972/1107706265898330/f

Fix the bug of retrieving wrong values for skipped frames and an idle crash related to calling its method.

### DIFF
--- a/js/module.d.ts
+++ b/js/module.d.ts
@@ -239,6 +239,7 @@ export declare const enum EOutputCode {
     NoSpace = -7
 }
 export declare const Global: IGlobal;
+export declare const Video: IVideo;
 export declare const OutputFactory: IOutputFactory;
 export declare const AudioEncoderFactory: IAudioEncoderFactory;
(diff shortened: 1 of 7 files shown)


summeroff (approved):

EddyGharbi (changes requested):
Good job finding this and fixing it! 

Only thing is that I'm a not a big fan of leaving deprecated code considering our past. If we're not careful we end up with a considerable amount of dead code within weeks. I would remove it and create a PR in the frontend along with the update of the module that uses the new API.

RodrigoHolztrattner:
Working on it!

RodrigoHolztrattner:
Changes done, frontend PR: https://github.com/stream-labs/streamlabs-obs/pull/1363

EddyGharbi (approved):
Ty, looks good 👍

PR #247, opened 2019-02-07
Handle the destruction of Projector Window (fix a crash)

RodrigoHolztrattner:
Asana task: https://app.asana.com/0/734357654754972/1107505184397281/f

The window HWND is attached to the projector HTML DOM, we cannot control its destruction so we must allow the `DestroyWindow()` method to fail with error 1400, invalid window reference, if it (HTML DOM) was able to destroy it first.

### DIFF
--- a/obs-studio-server/source/nodeobs_display.cpp
+++ b/obs-studio-server/source/nodeobs_display.cpp
@@ -115,22 +115,21 @@ struct DestroyWindowMessageQuestion
 struct DestroyWindowMessageAnswer : message_answer
 {};
 
-static void HandleWin32ErrorMessage()
+static void HandleWin32ErrorMessage(DWORD errorCode)
 {
-	DWORD dwErrorCode    = GetLastError();
 	LPSTR lpErrorStr     = nullptr;
 	DWORD dwErrorStrSize = 16;
 	DWORD dwErrorStrLen  = FormatMessageA(
         FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM,
         NULL,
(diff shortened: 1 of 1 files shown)


summeroff:
In case someone else try to close our window we can catch that be looking for messages WM_CLOSE and WM_DESTROY in OBS::Display::DisplayWndProc . And delete Display associated with destroyed HWND.  

It is more suggestion for an improvement as fix in this PR is ok and it small and fixes crash without increasing chances of side effects.

RodrigoHolztrattner:
Good suggestion! I will add it as a different task on Sentry since the reason for this PR is to fix the crash but would be nice to update the code to handle the way you said.

Update: Asana task: https://app.asana.com/0/734357654754972/1108635072226916/f

obs-studio-server/source/nodeobs_display.cpp
```diff
@@ -205,8 +204,19 @@ void OBS::Display::SystemWorker()
...
+                // We check for error 1400 because if this display is a projector, it is attached to a HTML DOM, so
+                // we cannot directly control its destruction since the HTML will probably do this concurrently, 
+                // the DestroyWindow is allows to fail on this case, a better solution here woul be checking if this
+                // display is really a projector and do not attempt to destroy it (let the HTML do it for us).
+				if (error != 1400) {
+					answer->success = false;
+					HandleWin32ErrorMessage(error);
+                } else {
+					answer->success = true;
+                }
```
EddyGharbi: Please fix indentation here

RodrigoHolztrattner: Done

PR #250, opened 2019-02-08
Fixes problems preventing tests running correctly

pdamasceno:
Fix 1: Make the pipe name being used in IPC.host() be unique by using uuid
Fix 2: Cast hotkeyId from string to number and initialize isKeyDown variable properly in process hotkey test case of nodeobs_api. Remove fail test of process hotkey, we don't check the id in the backend right now.
Fix 3: Remove fail test of find item in osn-scene test cases. I have a branch that reworks osn-scene tests in a way that the test cases are not dependant on each other, among other fixes related to osn-scene test cases.


EddyGharbi (approved):
Looks good to me 👍 Thanks for that

PR #252, opened 2019-02-08
Make shure we don't query a volmeter when it's being destroyed

RodrigoHolztrattner:
Found that would be good to have a mutex here to prevent querying an object that is being destroyed, this isn't causing any problem but it's better to have it to prevent any accidental assumption in the future.


RodrigoHolztrattner:
Actually I don't think this will be ever necessary, closing this PR

PR #256, opened 2019-02-11
Add remaining osn-cene test cases and rework test cases so that they don't depend on each other

pdamasceno:


EddyGharbi (approved):
Looks good to me

PR #258, opened 2019-02-13
Verify if the output channel is valid (less then the maximum allowed)

RodrigoHolztrattner:
This check is necessary since obs won't indicates an error if we pass an invalid channel.


EddyGharbi (approved):
Looks good to me 👍

PR #260, opened 2019-02-13
Update osn-video test cases to get values related to frames

pdamasceno:
* The way to get skipped frames and total frames (now encoded frames) values has changed


EddyGharbi (approved):
Looks good 👍

PR #262, opened 2019-02-13
Add osn-volmeter test cases

pdamasceno:


tests/osn-tests/src/test_osn_volmeter.ts
```diff
@@ -0,0 +1,93 @@
...
+            expect(input.name).to.equal('input');
+
+            // Creating volmeter
+            const volmeter = osn.VolmeterFactory.create(osn.EFaderType.IEC);
+
+            // Checking if volmeter was created correctly
+            expect(volmeter).to.not.equal(undefined);
+
+            // Attach volmeter to input source
+            volmeter.attach(input);
```
EddyGharbi: Is there a way we can if this worked?

tests/osn-tests/src/test_osn_volmeter.ts
```diff
@@ -0,0 +1,93 @@
...
+
+            // Adding callback to volmeter
+            const cb = volmeter.addCallback((magnitude: number[], peak: number[], inputPeak: number[]) => {});
+
+            // Checking if callback was added correctly
+            expect(cb).to.not.equal(undefined);
+
+            // Removing callback from volmeter
+            volmeter.removeCallback(cb);
```
EddyGharbi: Is there a way we can if this worked?

tests/osn-tests/src/test_osn_volmeter.ts
pdamasceno: Attach() does not return anything. Right now I can only check if the function is not returning an exception since ValidateResponse throws an exception if something went wrong. If we make the function return something, a boolean for example, I can check the return value after calling it.

tests/osn-tests/src/test_osn_volmeter.ts
pdamasceno: RemoveCallback actually returns Nan::Null() in case of failure or true in case of success. Totally missed this. I added a check of the return value to the test case.

EddyGharbi (approved):
Looks good to me 👍

PR #265, opened 2019-02-15
Fix an encoder race by not recreating the encoder if not necessary

RodrigoHolztrattner:
Asana task: https://app.asana.com/0/734357654754972/1109332300176239/f

Fix an thread race when starting streaming right after stop recording.

### DIFF
--- a/obs-studio-server/source/nodeobs_settings.cpp
+++ b/obs-studio-server/source/nodeobs_settings.cpp
@@ -1670,7 +1670,7 @@ SubCategory OBS_settings::getAdvancedOutputStreamingSettings(config_t* config, b
 	bool recStreamUsesSameEncoder   = streamingEncoder == recordEncoder;
 	bool recOutputBlockStreamOutput = !(!recStreamUsesSameEncoder || (recStreamUsesSameEncoder && !recOutputIsActive));
 
-	if (!streamOutputIsActive && !recOutputBlockStreamOutput && !fileExist) {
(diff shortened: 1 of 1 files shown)


RodrigoHolztrattner:
I'm kind worried about the `settings` variable here https://github.com/stream-labs/obs-studio-node/pull/265/files#diff-c0f970394d426a017d9b22a780e66234R1679

If for any reason it's null it could cause a crash. Is it guaranteed that the call to `obs_encoder_get_settings(streamingEncoder);` will always return a valid settings?

I think we never reached this case before because we were always recreating the encoder, but now it will only be recreated if the file doesn't exist (also those other conditionals there)

EddyGharbi:
I think your link is not pointing at the right thing.

RodrigoHolztrattner:
Oh I can't use a link from a line that isn't touched by the PR... well it is this settings that I'm talking about `settings         = obs_encoder_get_settings(streamingEncoder);` (right below my changes).
If the streaming encoder doesn't exist it will return a null settings, that would crash when used by other methods.

We could verify if it's null and not proceed, returning what is current on the `streamingSettings` variable (like we do if the `streamOutput` is null), would that be ok?

EddyGharbi:
I feel like it would be better to just check if the encoder is null here `if (!streamOutputIsActive && !recOutputBlockStreamOutput && !fileExist)` and if it is, then go into that condition to create a valid one. What do you think?

RodrigoHolztrattner:
That would be good, I will update with this check

EddyGharbi:
Looks good, merging 👍

PR #267, opened 2019-02-20
Make init API receives a version (string) parameter from the frontend

RodrigoHolztrattner:
Receives another string parameter that will be user for our backend crash reports to proper identify SLOBS version

### DIFF
--- a/obs-studio-client/source/nodeobs_api.cpp
+++ b/obs-studio-client/source/nodeobs_api.cpp
@@ -31,16 +31,18 @@ void api::OBS_API_initAPI(const v8::FunctionCallbackInfo<v8::Value>& args)
 {
 	std::string path;
 	std::string language;
+	std::string version;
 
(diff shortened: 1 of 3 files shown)


pdamasceno (approved):
Looks good to me.

EddyGharbi:
I would add using this string to associate the version number to the crash in this PR, hard to see the whole picture when reviewing without it.

RodrigoHolztrattner:
I would need to merge my Sentry PR again, or at least recreate the Crashpad instance to add more info.
We create the crashpad instance on the main function but we only receive the version from the init API (so, I would need to add some code from the init API to recreate it).
We could put this on hold until the next release with the backend changes to then integrate again the Sentry PR and start using this information, what approach you think would be better?

EddyGharbi:
Yes that sounds good 👍

RodrigoHolztrattner:
Sentry PR will add this change, closing this PR.

PR #268, opened 2019-02-20
Protect manager class from thread racing

RodrigoHolztrattner:
Asana task: https://app.asana.com/0/734357654754972/1110444842717820/f

This is necessary to prevent accessing the manager class methods from different threads at the same time, this would happen if `obs` internally decides to release a source (from its audio/video/graphics thread for ex) and at the same time we also release a source by ourselves (by changing the scene collection for ex), we could have 2 threads trying to remove a source that is inside a map and this potentially will crash at some time.

IMPORTANT: I would suggest to wait to merge this until the next next preview only since we already have too many additions right now.

### DIFF
--- a/obs-studio-server/source/utility.hpp
+++ b/obs-studio-server/source/utility.hpp
@@ -136,7 +136,7 @@ namespace utility
 
         void for_each(std::function<void(T*)> for_each_method)
         {
-			std::lock_guard<std::recursive_mutex> lock(internal_mutex);
+            std::lock_guard<std::recursive_mutex> lock(internal_mutex);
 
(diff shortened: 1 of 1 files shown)


summeroff (commented):
may be we need mutex in destructor too as it clears `object_map` as what `clear()` does. 
or just call clear() from destructor.

RodrigoHolztrattner:
Good point! I will add this tomorrow, thanks for the review!

RodrigoHolztrattner:
Done!

EddyGharbi (approved):
Looks good now 👍

PR #269, opened 2019-02-20
Add handle scope to callback_handler function to prevent Nan/V8 fatal error

pdamasceno:
When creating test cases for nodeobs_autoconfig I was running into the error below:
`FATAL ERROR: v8::HandleScope::CreateHandle() Cannot create handle without a HandleScope`

Node.js/Nan documentation says that you need to allocate a handle scope when creating v8 objects, so added a handle scope to callback_handler function to prevent the error above.

HandleScope documentation: https://github.com/nodejs/nan/blob/master/doc/scopes.md#api_nan_handle_scope


EddyGharbi:
Looks good 👍

EddyGharbi (approved):

PR #277, opened 2019-02-27
fix: check if config valid before use in api destroy

summeroff:
We have a lot of crashes lately on access to config value when in OBS_API::destroyOBS_API call (by info from sentry). 

Crash in `config_get_bool(ConfigManager::getInstance().getBasic(), "Audio", "DisableAudioDucking");`
It can be possible if something bad happen to `basic config` . From crash report it is not very clear that exactly but there some possible way where config not checked for null and goes to a function that can crash on that. 
 
Also while looking in code of `config_create` `config_open` functions. I find out that `config_open` also allocate memory for config and that does leaking memory allocated by config_create.


EddyGharbi (approved):
Looks good to me 👍

PR #280, opened 2019-02-28
Handle some crashes when loading modules

RodrigoHolztrattner:
Don't crash when receiving an exception if one of the loading or initialization methods failed.

Asana taks: https://app.asana.com/0/734357654754972/1110444842717822/f

### DIFF
--- a/obs-studio-server/source/nodeobs_api.cpp
+++ b/obs-studio-server/source/nodeobs_api.cpp
@@ -1141,7 +1141,17 @@ bool OBS_API::openAllModules(int& video_err)
 #endif
 
 			obs_module_t* module;
-			int           result = obs_open_module(&module, plugin_path.c_str(), plugin_data_path.c_str());
+			int           result;
+
+			try {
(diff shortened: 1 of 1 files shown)


obs-studio-server/source/nodeobs_api.cpp
```diff
@@ -1141,7 +1141,17 @@ bool OBS_API::openAllModules(int& video_err)
 #endif
 
 			obs_module_t* module;
-			int           result = obs_open_module(&module, plugin_path.c_str(), plugin_data_path.c_str());
+			int           result;
```
summeroff: it is better to init `result` with some default value. as if we catch exception no value be assigned to `result`.

RodrigoHolztrattner: Added initialization values for a few variables! Thanks for noticing!

summeroff (approved):

EddyGharbi (approved):

EddyGharbi (approved):
Looks good 👍

PR #282, opened 2019-03-02
Continue instead returning if a plugin path doesnt exist

RodrigoHolztrattner:
Asana task: https://app.asana.com/0/734357654754972/1107683930734842/f


EddyGharbi (approved):
Looks good 👍

PR #287, opened 2019-03-04
fix: change audio track names to conventional

summeroff:
change audio track names params to `TrackXName`. 
add code to move existing values to new names. 
fix mistype of Track5Bitrate.


EddyGharbi (approved):
Looks good to me 👍

PR #291, opened 2019-03-06
[WIP] Remove the parameter that would make the crashmanager to call abort after the crash

RodrigoHolztrattner:
This PR is an attempt to solve the wrong name the is being assigned to the reports on Sentry

For example:
`https://sentry.io/organizations/streamlabs-obs/issues/?project=1406061&statsPeriod=14d`


RodrigoHolztrattner:
Found a different approach to guarantee stacktrace rewind, this isn't necessary anymore

PR #292, opened 2019-03-06
Fix callstack rewind

RodrigoHolztrattner:
Fix the issue that sometimes there was no callstack rewind on the crash reports.
The library files added here should help on this task and also allow a few more functionalities that can be explored in the future.


RodrigoHolztrattner:
I added the license directly on the library files folder, since it's use the BSD-3 there is no problem on using the library. If I should put the license file on a different folder fell free to ping me.

EddyGharbi (approved):

PR #293, opened 2019-03-07
Fix sentry report names

RodrigoHolztrattner:
Fix the issue with the report names on Sentry.

The fix is to call the internal crashpad exception method manually when the windows `UnhandledExceptionFilter` method is called.


summeroff (approved):

EddyGharbi (approved):
Hard for me to have an idea if this is working correctly. It seems to look good.

PR #295, opened 2019-03-07
Disable the multitrack option if the recording format is flv

RodrigoHolztrattner:
Disable the multitrack option when the selected recording format is set to `flv`.
This will also prevent saving a different multitrack bitmask if the user had it enabled on another format and decides to change to `flv`. It will force a bitmask that activates only the first audio track.

### DIFF
--- a/obs-studio-server/source/nodeobs_audio_encoders.cpp
+++ b/obs-studio-server/source/nodeobs_audio_encoders.cpp
@@ -229,7 +229,7 @@ const char* GetAACEncoderForBitrate(int bitrate)
 
 bool IsMultitrackAudioSupported(const char* format)
 {
-	if (format != nullptr && strcmp(format, "flv") == 0) {
+	if (format == nullptr || strcmp(format, "flv") == 0) {
 		return false;
 	}
 
--- a/obs-studio-server/source/nodeobs_audio_encoders.cpp
+++ b/obs-studio-server/source/nodeobs_audio_encoders.cpp
@@ -227,6 +227,15 @@ const char* GetAACEncoderForBitrate(int bitrate)
 
 #define INVALID_BITRATE 10000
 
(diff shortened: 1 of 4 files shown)


summeroff (approved):

tests/osn-tests/src/test_nodeobs_settings.ts
```diff
@@ -1176,11 +1176,25 @@ describe('nodeobs_settings', function() {
...
+                }).parameters.find(parameter => {
+                    return parameter.name === 'RecFormat';
+                }).currentValue = selectedFormat;
+
                 osn.NodeObs.OBS_settings_saveSettings('Output', setCBR);
 
                 // Getting advanced output settings container with CBR parameters
                 let cbrOutputSettings = osn.NodeObs.OBS_settings_getSettings('Output');
-
+
```
EddyGharbi: Could you clean the spaces here?

obs-studio-server/source/nodeobs_audio_encoders.cpp
```diff
@@ -227,6 +227,15 @@ const char* GetAACEncoderForBitrate(int bitrate)
 
 #define INVALID_BITRATE 10000
 
+bool IsMultitrackAudioSupported(const char* format)
+{
+	if (format != nullptr && strcmp(format, "flv") == 0) {
```
EddyGharbi: if the format is null we should be returning false right?

RodrigoHolztrattner: Yes, fixed it

tests/osn-tests/src/test_nodeobs_settings.ts
RodrigoHolztrattner: Done

EddyGharbi (approved):
Looks good 👍

PR #296, opened 2019-03-08
Limit the maximum bitrate if not surround

RodrigoHolztrattner:
Limit the maximum bitrate to 320 if the current audio config isn't surround


summeroff (approved):

EddyGharbi (approved):
Looks good 👍

PR #324, opened 2019-03-18
Disable crash reporting before killing SLOBS process

RodrigoHolztrattner:
Client and Database doesn't have shutdown methods, since we are going to kill SLOBS process anyway we can call their destructors directly.

~Needs a few more testing before merge.~ Tested and ok!


EddyGharbi (approved):
Looks good 👍

PR #328, opened 2019-03-19
Make osn-tests easier to run locally and in appveyor

pdamasceno:
There is one change to the path of obs-studio-node in imports of all tests that had to be made in order to make it all work

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,12 +15,11 @@
   "main": "./index.js",
   "types": "./index.d.ts",
   "scripts": {
-    "clean": "git clean -fxd",
     "build:bindings": "cmake-js configure -a x64 --CDCMAKE_PREFIX_INSTALL=distribute && cmake --build build --target install --config Release",
     "build:javascript": "tsc -p js",
     "build:docs": "typedoc --out docs",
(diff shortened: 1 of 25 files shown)


package.json
```diff
@@ -18,14 +18,27 @@
     "clean": "git clean -fxd",
     "build:bindings": "cmake-js configure -a x64 --CDCMAKE_PREFIX_INSTALL=distribute && cmake --build build --target install --config Release",
     "build:javascript": "tsc -p js",
-    "build:docs": "typedoc --out docs"
+    "build:docs": "typedoc --out docs",
+    "local:config": "yarn install && git submodule update --init --recursive --force && cmake -Bbuild -H. -G\"Visual Studio 15 2017\" -A x64 -DCMAKE_INSTALL_PREFIX=\"./obs-studio-node\"",
+    "local:build": "cmake --build build --target install",
```
EddyGharbi: --config RelWithDebInfo/Debug is missing here.

pdamasceno: It uses Debug by default if --config is not used. Still want me to add it there?

EddyGharbi: I would make it explicit, doesn't really hurt and prevent some potential mistakes.

pdamasceno: Made it explicit in the latest commit

PR #330, opened 2019-03-19
Check if obs started correctly, if false log it

RodrigoHolztrattner:
Log if obs doesn't start correctly. 
The additional log is a measure to see if the `userdata` path is wrong and causing this issue https://sentry.io/organizations/streamlabs-obs/issues/944784869/?project=1283431&query=is%3Aunresolved&statsPeriod=14d&utc=false

When init API supports more return codes, would be a good idea to return an error code there (and on a few other places there too)


EddyGharbi (approved):
Looks good 👍

PR #331, opened 2019-03-19
Add stop signal code check to prevent test lockup

pdamasceno:
nodeobs_service could stay locked up after receiving a stop signal with a code different than 0


EddyGharbi (approved):
Looks good, although it would be nice to make sure we understand why it randomly fails.

PR #339, opened 2019-03-27
Recreate the service.json attributes if the file is blank/invalid

RodrigoHolztrattner:
Fix not being able to create a valid service object, resulting in a crash.


EddyGharbi (approved):
Looks good 👍

PR #340, opened 2019-03-27
Disable the reports if there are pending objects to be released on obs

RodrigoHolztrattner:
Verify if some sources are still alive on shutdown, meaning that the frontend (could have) crashed and it wasn't able to release all objects. If that happens `obs_shutdown()` have a huge change of crashing.

This PR will put that method inside a `try-catch` block and also disable the crash manager.


EddyGharbi (approved):
Looks good 👍

PR #342, opened 2019-04-03
Add the missing string include

RodrigoHolztrattner:
After migrating to VS2019 it started to detect this header was missing, not sure why VS17 didn't said anything about it.


EddyGharbi (approved):
Looks good 👍

PR #344, opened 2019-04-05
Add SceneItemData pointer nullptr check

pdamasceno:
After a source is deleted the GetId and SetSelected functions are called when you click SLOBS. This was causing a client crash because those functions were trying to access SceneItemData pointer of a scene item that don't exist anymore. Adding nullptr checks to these pointers in these functions resolves the crash.


EddyGharbi (approved):
Looks good 👍

PR #345, opened 2019-04-05
Fix replay buffer output signals

pdamasceno:
We were not sending any replay buffer output signals when recording path is invalid and output settings is in advanced mode

This also needs https://github.com/stream-labs/obs-studio/pull/55


EddyGharbi (approved):
Looks good 👍

PR #348, opened 2019-04-08
Add the ifdef guards

RodrigoHolztrattner:
Add the necessary `#ifdef` statements to allow debug builds without using the crashpad.


EddyGharbi (approved):
Looks good 👍

PR #355, opened 2019-04-16
Update osn-source tests

pdamasceno:
Updated the tests with real settings for sources, filters and transitions in the Update and Save functions tests
Removed the scene test case from Update and Save functions tests because scenes don't have settings


EddyGharbi (approved):
Looks good

PR #356, opened 2019-04-16
Fix node version error on osn-tests

pdamasceno:
- Change mocha to electron-mocha as test runner, this makes Mocha tests run in Electron. Now there are no more Node version errors when running tests with Node version above 8
- Upgraded Node version in appveyor.yml to version 10


summeroff (approved):

EddyGharbi (approved):
Looks good

PR #358, opened 2019-04-18
Add checks to prevent an invalid VA string from obs

RodrigoHolztrattner:
Fix sometimes receiving an error string with an invalid `variable arguments` set from obs.
When obs crash, sometimes it makes a call to `bcrash` that is linked to our crash manager object, sometimes obs could send an invalid VA parameter and this was causing a crash inside a crash (this can be seen on the report below)

https://sentry.io/organizations/streamlabs-obs/issues/990771384/?project=1283430&query=is%3Aunresolved

### DIFF
--- a/obs-studio-server/source/util-crashmanager.cpp
+++ b/obs-studio-server/source/util-crashmanager.cpp
@@ -215,7 +215,11 @@ bool util::CrashManager::Initialize()
 	// Handler for obs errors (mainly for bcrash() calls)
 	base_set_crash_handler(
 	    [](const char* format, va_list args, void* param) {
-		    std::string errorMessage = FormatVAString(format, args);
+		    std::string errorMessage;
+		    if (format == nullptr)
+			    errorMessage = "unknown error";
+		    else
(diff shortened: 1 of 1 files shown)


obs-studio-server/source/util-crashmanager.cpp
```diff
@@ -462,15 +466,21 @@ bool util::CrashManager::TryHandleCrash(std::string _format, std::string _crashM
 // Format a var arg string into a c++ std::string type
 std::string FormatVAString(const char* const format, va_list args)
 {
+	static const int MaximumVAStringSize = 63;
 	auto temp   = std::vector<char>{};
-	auto length = std::size_t{63};
+	auto             length              = std::size_t{MaximumVAStringSize};
```
EddyGharbi: Weird style here, could you fix?

RodrigoHolztrattner: Done

EddyGharbi (approved):
Looks good

PR #359, opened 2019-04-18
Do not recreate the service if the type is invalid

RodrigoHolztrattner:
Fix recreating the service if we detected that the type was null (basically we had a missing `else`, the service was being created using the null type and the crash wasn't prevented as it supposed to happen)


EddyGharbi (approved):
Looks good

PR #360, opened 2019-04-18
Adds a bool variable that will prevent from recreating the crashpad

RodrigoHolztrattner:
In some cases the client was being recreated even by reseting the older client and database objects, this boolean variable will prevent it from being recreated


EddyGharbi (approved):
Looks good

PR #361, opened 2019-04-22
Add warnings to our crash reports if the log file creation fails

RodrigoHolztrattner:
Add warnings to our crash report if the user is unable to create the log file (or its path).
Some crashes have empty log like this https://sentry.io/organizations/streamlabs-obs/issues/990898773/?project=1283431&query=is%3Aunresolved

Would be good to know why that is happening, the warning will also send the path that SLOBS attempted to create the log file


EddyGharbi (approved):
Looks good

PR #368, opened 2019-04-29
Electron 4

summeroff:
Electron 4 support

Also include fix for tests from PR https://github.com/stream-labs/obs-studio-node/pull/356


EddyGharbi (approved):
Looks good

PR #370, opened 2019-04-30
Setup backend metrics light

RodrigoHolztrattner:
Crash-handler PR: https://github.com/stream-labs/crash-handler/pull/7

This PR will make sure we will be sending periodic status updates to the crash-handler process indicating what method is being executed.

The remaining functionality is discussed on the crash-handler PR.


EddyGharbi (approved):
Looks good

PR #374, opened 2019-05-06
Fix tests not being able to run locally

pdamasceno:
The changes made to support electron 4 broke some stuff when running tests locally.
Added electron-rebuild package to rebuild NAN modules (osn on this case) whenever we do yarn install.

https://github.com/electron/electron/blob/v0.37.2/docs/tutorial/using-native-node-modules.md#how-to-install-native-modules


EddyGharbi:
Looks good

PR #377, opened 2019-05-07
Fix lossless recording with no audio on simple mode

pdamasceno:


obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -1604,6 +1604,8 @@ void OBS_service::LoadRecordingPreset_Lossless()
 	obs_data_set_string(settings, "video_encoder", "utvideo");
 	obs_data_set_string(settings, "audio_encoder", "pcm_s16le");
 
+   int aMixes = 1;
```
EddyGharbi: I see that this broke when I did the multi track audio recording. Thanks for fixing it, could you fix the style here?

EddyGharbi (approved):

PR #380, opened 2019-05-08
Checks the replay buffer output instead the recording one for errors

RodrigoHolztrattner:
This was the cause of a few crashes on shutdown and during idle (if the user was fast enough to start/stop recording + start/stop replay buffer)

We were using the recording output when retrieving the error since there was no check for the replay buffer type, the additions will handle this situation and prevent any future output that we decide to add from being accessed incorrectly if a entry of that type wasn't added on that `if`.

### DIFF
--- a/obs-studio-server/source/nodeobs_service.cpp
+++ b/obs-studio-server/source/nodeobs_service.cpp
@@ -2175,12 +2175,10 @@ void OBS_service::JSCallbackOutputSignal(void* data, calldata_t* params)
 			output = streamingOutput;
 		else if (signal.getOutputType().compare("recording") == 0)
 			output = recordingOutput;
-		else if (signal.getOutputType().compare("replay-buffer") == 0)
+		else (signal.getOutputType().compare("replay-buffer") == 0)
 			output = replayBufferOutput;
-		else
-			output = nullptr;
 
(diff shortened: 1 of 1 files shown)


obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -2173,10 +2173,14 @@ void OBS_service::JSCallbackOutputSignal(void* data, calldata_t* params)
...
+		else if (signal.getOutputType().compare("recording") == 0)
 			output = recordingOutput;
+		else if (signal.getOutputType().compare("replay-buffer") == 0)
+			output = replayBufferOutput;
+		else
+			output = nullptr;
 
-		const char* error = obs_output_get_last_error(output);
+		const char* error = (output ? obs_output_get_last_error(output) : "invalid output type");
 		if (error) {
```
EddyGharbi, line 2182: Seems like this check is not necessary anymore.

RodrigoHolztrattner: I added just in case we decide to add another output type without adding an if, but I will remove it to handle the cases that we have now.

RodrigoHolztrattner: Ready!

EddyGharbi (approved):
Ty! Looks good

PR #393, opened 2019-05-22
Release the advanced streaming audio encoder

RodrigoHolztrattner:
This could cause some crashes on idle and shutdown, possible also some inconsistencies when switching encoders

### DIFF
--- a/obs-studio-server/source/nodeobs_api.cpp
+++ b/obs-studio-server/source/nodeobs_api.cpp
@@ -1056,6 +1056,10 @@ void OBS_API::destroyOBS_API(void)
 	if (audioRecordingEncoder != NULL)
 		obs_encoder_release(audioRecordingEncoder);
 
+	obs_encoder_t* audioAdvancedStreamingEncoder = OBS_service::getAudioAdvancedStreamingEncoder();
+	if (audioAdvancedStreamingEncoder != NULL)
+		obs_encoder_release(audioAdvancedStreamingEncoder);
+
 	obs_output_t* streamingOutput = OBS_service::getStreamingOutput();
 	if (streamingOutput != NULL)
 		obs_output_release(streamingOutput);
--- a/obs-studio-server/source/nodeobs_service.cpp
+++ b/obs-studio-server/source/nodeobs_service.cpp
@@ -924,8 +924,10 @@ bool OBS_service::updateAudioStreamingEncoder() {
 	} else {
 		uint64_t trackIndex = config_get_int(ConfigManager::getInstance().getBasic(), "AdvOut", "TrackIndex");
 
-		if (audioAdvancedStreamingEncoder)
+		if (audioAdvancedStreamingEncoder) {
 			obs_encoder_release(audioAdvancedStreamingEncoder);
+			audioAdvancedStreamingEncoder = nullptr;
+		}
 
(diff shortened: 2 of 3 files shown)


obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -1973,6 +1977,17 @@ void OBS_service::setAudioSimpleRecordingEncoder(obs_encoder_t* encoder)
...
 }
 
+obs_encoder_t* OBS_service::getAudioAdvancedStreamingEncoder(void)
+{
+	return audioAdvancedStreamingEncoder;
+}
+
+void OBS_service::setAudioAdvancedStreamingEncoder(obs_encoder_t* encoder)
+{
+	obs_encoder_release(audioAdvancedStreamingEncoder);
```
summeroff, line 1989: does obs_encoder_release handle NULL correctly in case audioAdvancedStreamingEncoder is NULL?

RodrigoHolztrattner: Yes, actually I'm planning to remove all `if(encoder) release(encoder);` in the future, they are unnecessary since the function handles the null case.
Just gonna wait to remove they all together.

obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -924,8 +924,12 @@ bool OBS_service::updateAudioStreamingEncoder() {
 	} else {
 		uint64_t trackIndex = config_get_int(ConfigManager::getInstance().getBasic(), "AdvOut", "TrackIndex");
 
+		if (audioAdvancedStreamingEncoder)
```
summeroff: after obs_encoder_release encoder is not usable anymore, can a NULL be assigned to audioAdvancedStreamingEncoder?

RodrigoHolztrattner: It will be assigned in one of the lines below, but it's a good idea to make it null if we add more code here or modify the `if` in some way it doesn't assign a value, good point

obs-studio-server/source/nodeobs_api.cpp
```diff
@@ -1056,6 +1056,10 @@ void OBS_API::destroyOBS_API(void)
 	if (audioRecordingEncoder != NULL)
 		obs_encoder_release(audioRecordingEncoder);
 
+	obs_encoder_t* audioAdvancedStreamingEncoder = OBS_service::getAudioAdvancedStreamingEncoder();
```
summeroff, line 1059: can be replaced with _OBS_service::setAudioAdvancedStreamingEncoder( nullptr )_ 
it also will keep _audioAdvancedStreamingEncoder_  being changed only by its _setter_ function. 

also same point for another place where audioAdvancedStreamingEncoder being released.

RodrigoHolztrattner: I will make another PR when this is added to staging to address your suggestion, since until now we were releasing all of them this way I believe it's better to change they all together to make it consistent.
I believe a "ReleaseEncoders" method inside the `nodeobs_service` class will be better instead letting the `nodeobs_api` class decide if it should release them or not.

summeroff (approved):

EddyGharbi (approved):

PR #395, opened 2019-05-23
Monitor Memory

EddyGharbi:


pdamasceno (approved):

obs-studio-server/source/memory-manager.cpp
```diff
@@ -33,106 +32,248 @@ MemoryManager::MemoryManager()
...
+void MemoryManager::addCachedMemory(source_info* si)
+{
+	std::unique_lock<std::mutex> ulock(si->mtx);
+
+	if (!si->size || si->cached || current_cached_size + si->size > allowed_cached_size)
 		return;
 
-	obs_data_t* settings = obs_source_get_settings(source);
+	blog(LOG_INFO, "adding %dMB, source: %s", si->size / 1000000, obs_source_get_name(si->source));
+	current_cached_size += si->size;
```
RodrigoHolztrattner, line 113: If this can be called by multiple threads, shouldn't it be protected by a global mutex since it belongs to the memory manager singleton?
(same on the `removeCachedMemory`)

EddyGharbi: Seems like there is one missing in the sourceManager thread, but it is currently protected inside the monitorMemory thread, unregister function. Will add it there too 👍

RodrigoHolztrattner (approved):

PR #412, opened 2019-06-06
Prevent replay buffer from being stuck

RodrigoHolztrattner:
Add checks if it is already active and send a stop signal if it failed to start


summeroff (approved):

EddyGharbi (approved):
Looks good

PR #413, opened 2019-06-06
Add server throws

RodrigoHolztrattner:
Instead returning errors that would cause the client to throw an exception on the node side, this will make the server throw directly, potentially causing a crash report indicating correctly where the issue happened.
Until now it would cause the frontend to crash and we would not receive a report for it.

The `ValidateRensponse` on the client side must be kept since it still need to check if the `ipc` call was successful.

The `PRETTY_THROW` has its name because it uses `__PRETTY_FUNCTION__`.


EddyGharbi (approved):
Looks good

PR #439, opened 2019-07-08
Not crash in backend

summeroff:


obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -94,7 +94,8 @@ void OBS_service::OBS_service_resetAudioContext(
     std::vector<ipc::value>&       rval)
 {
 	if (!resetAudioContext(true)) {
-		PRETTY_THROW("OBS_service_resetAudioContext failed!");
+
```
EddyGharbi: Seems like there is an extra line here

obs-studio-server/source/osn-fader.cpp
```diff
@@ -78,12 +78,13 @@ void osn::Fader::Create(
...
 	obs_fader_t* fader = obs_fader_create(type);
 	if (!fader) {
-		PRETTY_THROW("failed to create");
+		PRETTY_ERROR_RETURN(ErrorCode::Error, "Failed to create Fader.");
 	}
 
 	auto uid = Manager::GetInstance().allocate(fader);
 	if (uid == std::numeric_limits<utility::unique_id::id_t>::max()) {
-		PRETTY_THROW("invalid uid");
+		obs_fader_destroy(fader);
```
EddyGharbi, line 86: Not sure I understand why we need to destroy the fader if we failed to allocate it

obs-studio-server/source/utility.hpp
```diff
@@ -31,15 +31,17 @@
...
-#define PRETTY_THROW(_message)                                                                 \
+#define PRETTY_ERROR_RETURN(_error_code, _message)                                                    \
 {                                                                                              \
 if (utility::osn_current_version() == "0.00.00-preview.0") {                                   \
     rval.push_back((uint64_t)ErrorCode::Error);                                                \
     return;                                                                                    \
 } else {                                                                                       \
-    auto error_message = std::string(__PRETTY_FUNCTION__) + " " + std::string(_message);       \
+    rval.push_back(ipc::value((uint64_t)_error_code));	    	                               \
+	rval.push_back(ipc::value(_message));				                                       \
```
EddyGharbi: Would be great to fix the style here

obs-studio-server/source/osn-fader.cpp
summeroff: This line from older code before PRETTY_THROW macros. 
As I understand In this case `obs_fader_create` was successful and then `allocate` failed we destroy what was created by `obs_fader_create` before return error to frontend.

EddyGharbi: kk I see it now, sounds good 👍

EddyGharbi (approved):

PR #457, opened 2019-08-09
Fix volmeter being stuck after changing output/input device or hiding source

pdamasceno:


obs-studio-server/source/osn-volmeter.cpp
```diff
@@ -314,3 +320,21 @@ void osn::VolMeter::OBSCallback(
...
+	auto currentTimeMs = std::chrono::time_point_cast<std::chrono::milliseconds>(currentTime);
+	return currentTimeMs.time_since_epoch();
+}
+
+bool osn::VolMeter::CheckIdle(std::chrono::milliseconds currentTime, std::chrono::milliseconds lastUpdateTime)
+{
+	auto idleTime = std::chrono::duration_cast<std::chrono::milliseconds>(currentTime - lastUpdateTime);
+
+	if (lastUpdateTime != std::chrono::milliseconds(0) && idleTime > std::chrono::milliseconds(300)) {
+		return true;
```
summeroff, line 339: You can split this IF in two. First check only lastUpdateTime to equal zero ms. And only if it is not then call GetTime(), calculate idleTime and compare it to 300 ms.  
This way cpu will not waste time on GetTime() and idleTime when they does not needed.

summeroff (approved):

EddyGharbi (approved):

PR #459, opened 2019-08-12
Add browser source interaction

EddyGharbi:


obs-studio-server/source/osn-source.cpp
```diff
@@ -734,6 +775,131 @@ void osn::Source::SetEnabled(
...
+
+	if (src == nullptr) {
+		rval.push_back(ipc::value((uint64_t)ErrorCode::InvalidReference));
+		rval.push_back(ipc::value("Source reference is not valid."));
+		AUTO_DEBUG;
+		return;
+	}
+
+	char* text = new char[args[2].value_str.size()];
+	strcpy(text, args[2].value_str.c_str());
```
summeroff, line 889: isnt it need to be size()+1 to copy zero character at the end of the c string?
can we send "args[2].value_str.c_str()" without creating new string?

obs-studio-server/source/osn-source.cpp
```diff
@@ -154,6 +154,47 @@ void osn::Source::Register(ipc::server& srv)
...
+	                           ipc::type::Int32,
+	                           ipc::type::UInt32},
+	    SendMouseClick));
+	cls->register_function(std::make_shared<ipc::function>(
+	    "SendMouseMove",
+	    std::vector<ipc::type>{ipc::type::UInt64,
+	                           ipc::type::UInt32,
+	                           ipc::type::UInt32,
+	                           ipc::type::UInt32,
+                               ipc::type::Int32},
```
pdamasceno: Indentation is different here

EddyGharbi: Fixed, ty, good catch

obs-studio-server/source/osn-source.cpp
EddyGharbi: No `obs_key_event` only takes a `char*` not `const char*`
Fixing the size, good catch

summeroff (approved):

pdamasceno (approved):

PR #477, opened 2019-09-05
Smart recording rebase

EddyGharbi:


obs-studio-client/source/sceneitem.cpp
```diff
@@ -265,6 +267,136 @@ Nan::NAN_METHOD_RETURN_TYPE osn::SceneItem::SetSelected(Nan::NAN_METHOD_ARGS_TYP
...
+		return;
+
+	std::vector<ipc::value> response =
+	    conn->call_synchronous_helper("SceneItem", "IsStreamVisible", std::vector<ipc::value>{ipc::value(item->itemId)});
+
+	if (!ValidateResponse(response))
+		return;
+	bool streamVisible = !!response[1].value_union.ui32;
+
+	sid->streamVisibleChanged = false;
```
pdamasceno: Best to verify if `sid` is valid before changing its values

summeroff (approved):

obs-studio-client/source/sceneitem.cpp
```diff
@@ -265,6 +267,136 @@ Nan::NAN_METHOD_RETURN_TYPE osn::SceneItem::SetSelected(Nan::NAN_METHOD_ARGS_TYP
...
+		return;
+
+	std::vector<ipc::value> response = conn->call_synchronous_helper(
+	    "SceneItem", "IsRecordingVisible", std::vector<ipc::value>{ipc::value(item->itemId)});
+
+	if (!ValidateResponse(response))
+		return;
+	bool recordingVisible = !!response[1].value_union.ui32;
+
+	sid->recordingVisibleChanged = false;
```
pdamasceno, line 363: Same thing here, verify if `sid` is valid before changing its values

obs-studio-client/source/sceneitem.cpp
EddyGharbi: Good catch! Fixed 👍

obs-studio-client/source/sceneitem.cpp
EddyGharbi: Good catch! Fixed 👍

pdamasceno (approved):

PR #483, opened 2019-09-13
Fix filter order reset

pdamasceno:
We were not updating the order in the filters container in the cache manager
The first time we get the filters from a source works fine because we get them directly from libobs (sdi->filtersOrderChanged = true)
Subsequent calls to get filters start getting them from the cache manager and there the order was not right

### DIFF
--- a/obs-studio-client/source/input.cpp
+++ b/obs-studio-client/source/input.cpp
@@ -918,29 +918,29 @@ Nan::NAN_METHOD_RETURN_TYPE osn::Input::SetFilterOrder(Nan::NAN_METHOD_ARGS_TYPE
 				if (it != sdi->filters->begin()) {
 					std::iter_swap(it, it - 1);
 				}
-		    } break;
+			} break;
 
 			case FilterMovement::DOWN: {
-			    if (it != std::prev(sdi->filters->end())) {
-				    std::iter_swap(it, it + 1);
-			    }
-		    } break;
+				if (it != std::prev(sdi->filters->end())) {
+					std::iter_swap(it, it + 1);
+				}
+			} break;
 
 			case FilterMovement::TOP: {
-			    auto rit = std::make_reverse_iterator(it);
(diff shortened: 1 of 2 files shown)


obs-studio-client/source/input.hpp
```diff
@@ -26,6 +26,14 @@ namespace osn
 {
 	class Input : public osn::ISource, public utilv8::ManagedObject<osn::Input>
 	{
+		enum FilterMovement
```
summeroff: int values of this enums have fixed values on frontend. 
So better to have them explicitly set in this enum too and not expect that compiler always generate them exactly as we need.

pdamasceno: Ok! Making the change

obs-studio-client/source/input.cpp
```diff
@@ -908,6 +910,30 @@ Nan::NAN_METHOD_RETURN_TYPE osn::Input::SetFilterOrder(Nan::NAN_METHOD_ARGS_TYPE
 
 	SourceDataInfo* sdi = CacheManager<SourceDataInfo*>::getInstance().Retrieve(baseobj->sourceId);
 	if (sdi) {
+		// Update order on cache manager filter container
+		auto it = std::find(sdi->filters->begin(), sdi->filters->end(), basefilter->sourceId);
+
+		if (movement == FilterMovement::UP && it != sdi->filters->begin()) {
+			std::iter_swap(it, it - 1);
+		}
```
summeroff: This 4 if's can be joined by "else" or replaced by "switch" as they expected to work exclusively. If one of them had worked no need to check another.

pdamasceno: You are right. I'll change it to a switch

summeroff (approved):

summeroff (approved):

EddyGharbi (approved):

summeroff (approved):

PR #484, opened 2019-09-17
Smart recording rebase outputs logic

EddyGharbi:


obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -911,110 +893,230 @@ bool OBS_service::startStreaming(void)
...
 	}
 
-	if (!codec) {
-		return false;
+	obs_encoder_t** enc = isSimpleMode ? &audioSimpleStreamingEncoder : &audioAdvancedStreamingEncoder;
+
+	if (*enc && obs_encoder_active(*enc))
+		return;
+
+	if (*enc && isSimpleMode) {
```
summeroff, line 937: If this condition works we have *enc released and set to null. 
In next condition "**(strcmp(codec, "aac") == 0 && isSimpleMode)**" we did not recreate this reference but use it 18 lines below in "obs_encoder_set_audio(*enc, obs_get_audio());"

And in another condition we check "**(strcmp(codec, "aac") == 0 && !isSimpleMode)**" and save to *enc another value. can it be a leak of encoder?

summeroff (approved):

obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -2122,16 +1979,11 @@ void OBS_service::setReplayBufferOutput(obs_output_t* output)
 	replayBufferOutput = output;
 }
 
-void OBS_service::updateStreamSettings(void)
+void OBS_service::updateStreamingOutput(bool isSimpleMode)
 {
 	const char* currentOutputMode = config_get_string(ConfigManager::getInstance().getBasic(), "Output", "Mode");
```
pdamasceno, line 1983: We are passing `isSimpleMode` to this function but it is never used. Instead of getting the output mode again it is better to use `isSimpleMode` in the `if` right below this

obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -799,7 +771,7 @@ bool OBS_service::createService()
 		obs_data_set_obj(data, "settings", settings);
 	};
 
-    bool fileExist = (os_stat(ConfigManager::getInstance().getService().c_str(), &buffer) == 0);
+	bool fileExist = (os_stat(ConfigManager::getInstance().getService().c_str(), &buffer) == 0);
```
pdamasceno: Different code style here

obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -825,7 +797,7 @@ bool OBS_service::createService()
 				blog(LOG_WARNING, "Failed to retrieve a valid service type from the data, using default properties!");
 				CreateNewService();
 
-			// Create the service normally since the service.json info looks valid
+				// Create the service normally since the service.json info looks valid
```
pdamasceno: Different code style here

obs-studio-server/source/nodeobs_service.cpp
EddyGharbi: Fixed 👍

obs-studio-server/source/nodeobs_service.cpp
EddyGharbi: Fixed 👍

obs-studio-server/source/nodeobs_service.cpp
EddyGharbi: Fixed 👍

PR #513, opened 2019-10-10
Fix removal of cache files in tests

pdamasceno:


tests/osn-tests/util/general.ts
```diff
@@ -41,6 +41,8 @@ export {game_captureHotkeys};
 let dshow_wasapitHotkeys: string[] = ['MUTE', 'UNMUTE', 'PUSH_TO_MUTE', 'PUSH_TO_TALK'];
 export {dshow_wasapitHotkeys};
 
+let deleteTries = 3;
```
EddyGharbi: This variable doesn't seem to be used.

EddyGharbi (approved):

PR #532, opened 2019-10-28
look log to detect ffmpeg error

summeroff:
Filter log messages to detect lines from a ffmpeg outdated drivers error. 
Enable filtering only for time when output started. 

Need this https://github.com/stream-labs/obs-studio/pull/157 to detect error.

### DIFF
--- a/obs-studio-server/source/nodeobs_api.cpp
+++ b/obs-studio-server/source/nodeobs_api.cpp
@@ -267,6 +267,65 @@ static void DeleteOldestFile(const char* location, unsigned maxLogs)
 #include <unistd.h>
 #endif
 
+outdated_driver_error * outdated_driver_error::inst = nullptr;
+
+outdated_driver_error * outdated_driver_error::instance()
+{
+	if(inst == nullptr)
+	{
+		inst = new outdated_driver_error();
+	}
+	return inst;
+}
+
+void outdated_driver_error::set_active( bool state) 
+{
+	if (state) {
+		if (!lookup_enabled)
+		{
+			line_1 = "";
+			line_2 = "";
+		}
+		lookup_enabled++;
+	} else {
+		lookup_enabled--;
+	}
+}
+
+std::string outdated_driver_error::get_error(const char* base_error)
+{
+	if (std::string(base_error).find("Function not implemented") == std::string::npos)
+		return "";
+
+	if(line_1.size() && line_2.size())
+		return line_1 + std::string("\n") + line_2;
+	else 
+		return "";
+}
+
+void outdated_driver_error::catch_error(const char* msg)
+{
+	if (!lookup_enabled)
+		return;
+
+	const std::string msg_string = msg;
+
+	if (line_1.size()==0) {
+		const std::string line_1_pattern = "Driver does not support the required nvenc API version";
+		auto pattern_position = msg_string.find(line_1_pattern);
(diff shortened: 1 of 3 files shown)


pdamasceno (approved):

obs-studio-server/source/nodeobs_api.cpp
```diff
@@ -267,6 +267,65 @@ static void DeleteOldestFile(const char* location, unsigned maxLogs)
...
+		}
+		lookup_enabled++;
+	} else {
+		lookup_enabled--;
+	}
+}
+
+std::string outdated_driver_error::get_error(const char* base_error)
+{
+	if (std::string(base_error).find("Function not implemented") == std::string::npos)
```
EddyGharbi: I would add a check here since base_error can be null

obs-studio-server/source/nodeobs_api.h
```diff
@@ -158,3 +158,18 @@ class OBS_API
 	static void SetProcessPriority(const char* priority);
 	static void destroyOBS_API(void);
 };
+
+class outdated_driver_error 
+{
+	static outdated_driver_error * inst;
+	std::string line_1 = ""; 
+	std::string line_2 = "";
+	int lookup_enabled = 0;
```
EddyGharbi, line 167: this variable is meant to be a bool no?

obs-studio-server/source/nodeobs_api.cpp
```diff
@@ -267,6 +267,65 @@ static void DeleteOldestFile(const char* location, unsigned maxLogs)
...
+
+void outdated_driver_error::catch_error(const char* msg)
+{
+	if (!lookup_enabled)
+		return;
+
+	const std::string msg_string = msg;
+
+	if (line_1.size()==0) {
+		const std::string line_1_pattern = "Driver does not support the required nvenc API version";
```
EddyGharbi, line 315: I would define this string in a #define, same for `const std::string line_2_pattern`

obs-studio-server/source/nodeobs_api.h
summeroff: No, as it is used as a counter. 
For example if two threads requested to enable lookup for driver error. It will be disabled only after two request for disabling be received.

EddyGharbi: I see, how can this happen though? All of this logic is in the main thread no?

obs-studio-server/source/nodeobs_api.cpp
summeroff: I have thought of making this two variables explicitly static. To tell a compiler not to recreate it on every function call in case it does not do so based only on const.  
Converting to a #define will make them more loosely typed and possibly add a type conversion.

obs-studio-server/source/nodeobs_api.h
summeroff: Yes. For this exact error we receive it immediately after obs_output_start() call. And as it only in our thread it cannot overlap. If it changed or we use it for other ffmpeg errors code will work without change. 
Complexity of this flag-counter is contained inside outdated_driver_error class and not making outside code more complicated.

obs-studio-server/source/nodeobs_api.cpp
summeroff: check added.

EddyGharbi (approved):

PR #545, opened 2019-11-06
Isolate test cases and abstract necessary code

pdamasceno:


tests/osn-tests/util/transition_settings.ts
```diff
@@ -0,0 +1,16 @@
+import { ISettings } from '../osn';
+
+// Transition settings
+
+let fadeToColor: ISettings = {
+    color: 4278190080,
```
summeroff, line 6: Replacing 4278190080 with 0xFF000000 will make it more easily distinguishable.

EddyGharbi (approved):

PR #591, opened 2020-01-15
Rename services.ts to better represent what it does

pdamasceno:


tests/osn-tests/util/user_pool_handler.ts
```diff
@@ -19,10 +19,9 @@ interface ITestUser {
     streamKey: string;
 }
 
-export class Services {
+export class UserPoolHandler {
     private user: ITestUser;
     private userPoolUrl: string = 'https://slobs-users-pool.herokuapp.com/';
-    private clientId = '8bmp6j83z5w4mepq0dn0q1a7g186azi';
```
EddyGharbi, line 25: what was this token? Let's make sure all access is revoked

pdamasceno: It was a token to request users Twitch users from the user-pool. We had to validate it in order get users. Alex did a change to the user-pool last year that the validation is now happening inside his user-pool code. I did the changes to reflect that on our side but forgot to remove this unused token. That token is not available anymore

EddyGharbi: Ok, sounds good, let's make sure going forward no token is visible in our code.

EddyGharbi (approved):

PR #607, opened 2020-02-11
Additional performance statistics

pdamasceno:


summeroff (approved):

obs-studio-client/source/nodeobs_api.cpp
```diff
@@ -89,11 +89,29 @@ void api::OBS_API_getPerformanceStatistics(const v8::FunctionCallbackInfo<v8::Va
 	    v8::String::NewFromUtf8(args.GetIsolate(), "percentageDroppedFrames").ToLocalChecked(),
 	    v8::Number::New(args.GetIsolate(), response[3].value_union.fp64));
 	statistics->Set(
-	    v8::String::NewFromUtf8(args.GetIsolate(), "bandwidth").ToLocalChecked(),
+	    v8::String::NewFromUtf8(args.GetIsolate(), "streamingBandwidth").ToLocalChecked(),
```
EddyGharbi: This is a breaking change streamlabs-obs client should integrate as part of this release.

obs-studio-server/source/nodeobs_api.cpp
```diff
@@ -1478,18 +1495,69 @@ double OBS_API::getCurrentBandwidth(void)
...
+double OBS_API::getAverageTimeToRenderFrame()
+{
+	return (double)obs_get_average_frame_time_ns() / 1000000.0;
+}
+
+std::string OBS_API::getDiskSpaceAvailable()
+{
+#define MBYTE (1024ULL * 1024ULL)
+#define GBYTE (1024ULL * 1024ULL * 1024ULL)
+#define TBYTE (1024ULL * 1024ULL * 1024ULL * 1024ULL)
```
EddyGharbi: We shouldn't re define these every time a call to `getDiskSpaceAvailable()` is made.

obs-studio-server/source/nodeobs_api.cpp
```diff
@@ -1453,14 +1464,20 @@ double OBS_API::getDroppedFramesPercentage(void)
 	return percent;
 }
 
-double OBS_API::getCurrentBandwidth(void)
+void OBS_API::OutputStats::getCurrentStats(bool recording)
```
EddyGharbi: In my opinion it would be easier to return `kbitsPerSec` and `dataOutput` here rather than using global variables.

obs-studio-client/source/nodeobs_api.cpp
pdamasceno: So get it back to `bandwidth` then?

obs-studio-server/source/nodeobs_api.cpp
pdamasceno: Will move those defines outside of there

obs-studio-client/source/nodeobs_api.cpp
EddyGharbi: We can totally change it, we just need to correctly apply the change as part of the PR that will update the osn in the frontend.

obs-studio-server/source/nodeobs_api.cpp
pdamasceno: I would like to keep it this way because the storing and calculations of `kbitsPerSec` and `dataOuput` are all handled by the structure `OutputStats`. It is better organized this way to have just two objects: one for the streaming output and another for the recording output. I also need to keep track of `lastBytesSent` and `lastBytesSentTime` to correctly calculate the values of `kbitsPerSec` and `dataOutput`, so it is either 2 global variables or 4 more variables in the `OBS_API` class, 2 for each output

obs-studio-server/source/nodeobs_api.cpp
pdamasceno: Moved those defines outside of the function

obs-studio-server/source/nodeobs_api.cpp
EddyGharbi: Ok I see, I would do 2 things then:
 1. Remove `getCurrentStats` from being a function inside the `OutputStats`, would be clearer to put it as a static function.
 2. Remove `bool recording` is the first argument and put `obs_output_t *output` instead. Then inside the function, make sure the output is not null.

pdamasceno: Sure thing, will get to make those changes

pdamasceno: Changes done. I also renamed `getCurrentStats` to `getCurrentOutputStats`

EddyGharbi (approved):

PR #608, opened 2020-02-14
use local scale to make autofit in auto gamecapture mode

summeroff:
need https://github.com/stream-labs/obs-studio/pull/174


EddyGharbi (approved):

EddyGharbi:
Will merge once OBS update is in.

PR #619, opened 2020-03-26
add option to use args from launcher instead of api init call

summeroff:


viral98:
Any ETA on when this will be merged?

EddyGharbi:
Closing this PR, thank you for your contribution and we apologize for the delay in replying. We ultimately decided to not move forward with the launcher approach, it was showing some promises but the risk of introducing major regressions to our users was to great.

PR #622, opened 2020-04-13
Potential fix with rtmp_common_key while start streaming

borzun:
- Restrict saving stream settings if stream is in progress;
- Undefined behavior fix about double deletion of `hotkeydata`.


borzun:
Surpassed by fix in `streamlabs-obs` project

PR #624, opened 2020-04-20
Fix data race when destroy streamingOutput

borzun:
- fix data race when destroy `streamingOutput` with simultaneous starting of the stream;
- add additional logs to track crashes when recreating service in`OBS_settings::saveStreamSettings` call;
- fix undefined behavior when double-delete `hotkeyData`.


borzun:
Closed due to problems with forked repository

PR #626, opened 2020-04-23
Fix data race when destroy streamingOutput

borzun:
- fix data race when destroy `streamingOutput` with simultaneous starting of the stream;
- add additional logs to track crashes when recreating service in`OBS_settings::saveStreamSettings` call;
- fix undefined behavior when double-delete `hotkeyData`.

### DIFF
--- a/obs-studio-server/source/nodeobs_service.cpp
+++ b/obs-studio-server/source/nodeobs_service.cpp
@@ -879,6 +879,8 @@ bool OBS_service::startStreaming(void)
 	if (!type)
 		type = "rtmp_output";
 
+	waitReleaseWorker();
+
 	if (streamingOutput)
 		obs_output_release(streamingOutput);
 
@@ -1148,7 +1150,8 @@ void OBS_service::stopStreaming(bool forceStop)
 
 	waitReleaseWorker();
 
-	releaseWorker = std::thread(releaseStreamingOutput);
+	releaseWorker   = std::thread(releaseStreamingOutput, streamingOutput);
+	streamingOutput = nullptr;
 
 	isStreaming = false;
 }
@@ -2255,17 +2258,16 @@ void OBS_service::duplicate_encoder(obs_encoder_t** dst, obs_encoder_t* src, uin
 	}
 }
 
-void OBS_service::releaseStreamingOutput()
+void OBS_service::releaseStreamingOutput(obs_output_t* output)
 {
 	if (config_get_bool(ConfigManager::getInstance().getBasic(), "Output", "DelayEnable")) {
-		uint32_t delay = obs_output_get_active_delay(streamingOutput);
+		uint32_t delay = obs_output_get_active_delay(output);
 		while (delay != 0) {
-			delay = obs_output_get_active_delay(streamingOutput);
+			delay = obs_output_get_active_delay(output);
 		}
 	}
-	
-	obs_output_release(streamingOutput);
-	streamingOutput = nullptr;
+
+	obs_output_release(output);
 }
 
 void OBS_service::waitReleaseWorker()
--- a/obs-studio-server/source/nodeobs_service.cpp
+++ b/obs-studio-server/source/nodeobs_service.cpp
@@ -2265,6 +2265,9 @@ void OBS_service::duplicate_encoder(obs_encoder_t** dst, obs_encoder_t* src, uin
 
 void OBS_service::releaseOutputWithActiveDelay(obs_output_t* output)
 {
+	if (!output)
+		return;
+
 	uint32_t delay = obs_output_get_active_delay(output);
 	while (delay != 0) {
 		delay = obs_output_get_active_delay(output);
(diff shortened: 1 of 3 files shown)


obs-studio-server/source/nodeobs_settings.cpp
```diff
@@ -1010,6 +1010,13 @@ void OBS_settings::saveStreamSettings(std::vector<SubCategory> streamSettings)
 
 	obs_data_release(hotkeyData);
 
+	if (OBS_service::isStreamingOutputActive()) {
+		blog(
+		    LOG_WARNING,
+		    "Do not create a new service while old '%s' service is active and streaming",
+		    currentServiceName);
```
EddyGharbi: Should we do more than just adding a log here?

obs-studio-server/source/nodeobs_service.h
```diff
@@ -176,7 +176,7 @@ class OBS_service
 	static void stopReplayBuffer(bool forceStop);
 	static void stopRecording(void);
 
-	static void releaseStreamingOutput(void);
+	static void releaseStreamingOutput(obs_output_t* output);
```
EddyGharbi: Very odd that `releaseStreamingOutput` takes an `output` as an argument, I suggest we review the logic here since it is starting to get very wrong.

obs-studio-server/source/nodeobs_settings.cpp
borzun: I added logs as a 'downgrade' from this PR: https://github.com/stream-labs/obs-studio-node/pull/622 ;

The problem was here is that frontend invoked `saveStreamSettings `during streaming, which is incorrect (the fix is here - https://github.com/stream-labs/streamlabs-obs/pull/2447 ). Thus, this log was added to track incorrect usage of backend when this crash will happen again.

EddyGharbi: I see, there should be no issue if we also return in the backend in case a call is indeed made from the frontend while streaming?

obs-studio-server/source/nodeobs_service.h
borzun: Before my changes we had a data race, because the `streamingOutput` could be changed from 2 threads. This is  wrong.

My changes basically eliminates a data race, because the `releaseStreamingOutput `doesn't need to access `streamingOutput` global in order to delete it. And when `releaseStreamingOutput` is invoked, no one should store the output, cause it will be eventually deleted.

obs-studio-server/source/nodeobs_settings.cpp
borzun: from my perspective - there are shouldn't be any issue. However, as I remember, @summeroff was against this option to silently return.

EddyGharbi: In my opinion, we can't just return, we had to fix it in the frontend since the logic was wrong there. Now nothing prevents us to make sure our code doesn't crash if the frontend starts re doing the same thing in the future.

summeroff: I think that it is better to return error to frontend where it can be recognized and fixed, waiting for crash reports will make it longer and user experience worse. 

It can be done in separate PR as it is not a main issue of this PR.

obs-studio-server/source/nodeobs_service.h
EddyGharbi: Sounds good for the explanation. Please re work on the naming.

obs-studio-server/source/nodeobs_settings.cpp
EddyGharbi: Agree, I would like to see the returning as part of this PR as a first step. Please make the change.

borzun: Added return. Also, moved this guard to the top of the function to protect from torn state of stream settings:

EddyGharbi: Perfect

obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -2271,17 +2279,14 @@ void OBS_service::duplicate_encoder(obs_encoder_t** dst, obs_encoder_t* src, uin
...
 
-void OBS_service::releaseStreamingOutput()
+void OBS_service::releaseOutputWithActiveDelay(obs_output_t* output)
 {
-	if (config_get_bool(ConfigManager::getInstance().getBasic(), "Output", "DelayEnable")) {
-		uint32_t delay = obs_output_get_active_delay(streamingOutput);
-		while (delay != 0) {
-			delay = obs_output_get_active_delay(streamingOutput);
-		}
+	uint32_t delay = obs_output_get_active_delay(output);
```
EddyGharbi, line 2271: I would add a check to at least make sure the output is non-null.

borzun: Done

EddyGharbi (approved):

PR #628, opened 2020-04-28
Fix incorrect device size after device rebuild

borzun:
After the device is rebuilt, need to update the device size, because sometimes it might be invalid. Thus, need to listen for device rebuild callback


borzun:
Cancelled jobs due to changes in the OBS API, which need to be merged first (see  stream-labs/obs-studio#187 )

summeroff (approved):

EddyGharbi (approved):

EddyGharbi:
Changes in OBS has been merged, will merge once OBS get merged in OSN.

PR #630, opened 2020-04-29
Upgrade cpp standard and fix compilcation error

EddyGharbi:


summeroff (approved):

pdamasceno (approved):

obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -685,7 +685,7 @@ static void ensure_directory_exists(std::string& path)
 
 	std::string directory = path.substr(0, last);
 
-	if (std::experimental::filesystem::is_directory(directory))
+	if (std::filesystem::is_directory(directory))
```
summeroff: isnt it should create dirs if is_directory is false?

EddyGharbi: Correct, fixing, this must have been broken for a long time. It seems like the behavior was the same with the old API.

PR #658, opened 2020-05-28
Added docs

### LINKED ISSUE #578: Docs / Examples avalible?
Is there any documentation or examples avalible? I'd like to use this in an electron project. I had a look at the TypeScript definitions and at the Stream-labs obs code, but since this looks very complicated to a developer without vue experience, I couldn't even figure out how to mix two sources together and get the result.

hrueger:
This should fix #578


hrueger:
I don't think that the checks are failing because of my changes...

PR #659, opened 2020-05-28
Add sanity checks when getting hotkey info

EddyGharbi:


obs-studio-server/source/nodeobs_api.cpp
```diff
@@ -868,8 +868,17 @@ void OBS_API::QueryHotkeys(
 		    }
 
 		    // Key defs
-		    auto       key_name = std::string(obs_hotkey_get_name(key));
-		    auto       desc     = std::string(obs_hotkey_get_description(key));
+		    const char* _key_name = obs_hotkey_get_name(key);
+		    const char* _desc     = obs_hotkey_get_description(key);
+
+		    if (!_key_name)
```
summeroff, line 874: may be better to skip key what does not have a name?

EddyGharbi: Good point, updated 👍

PR #677, opened 2020-06-16
Electron 903

summeroff:


summeroff:
Node versions compatibility was resolved in other way by moving to naip. Changes are in a main branch now.

PR #731, opened 2020-07-31
fix obs crash id sent to appState

summeroff:
I have tested by introducing a crashes and long sleep into server functions called sync and async.

### DIFF
--- a/obs-studio-server/source/util-crashmanager.cpp
+++ b/obs-studio-server/source/util-crashmanager.cpp
@@ -1046,7 +1046,7 @@ void util::CrashManager::SaveToAppStateFile()
 	if (current_status.size() == 0) 
 		return;
 	
-	const std::string freez_flag = "ipc_freez";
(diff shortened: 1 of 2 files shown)


obs-studio-server/source/util-crashmanager.cpp
```diff
@@ -1046,7 +1046,7 @@ void util::CrashManager::SaveToAppStateFile()
 	if (current_status.size() == 0) 
 		return;
 	
-	const std::string freez_flag = "ipc_freez";
+	const std::string freez_flag = "obs_crash";
```
EddyGharbi: I don't understand the logic behind the naming here, could you explain why the freeze flag is considered as an obs crash?

summeroff: Good point. 
Changed var name to not include mention of exact value.

EddyGharbi:
Why is the IPC diff so big? Could you make sure to be on the latest commit?

PR #793, opened 2020-11-25
Version check between client and server

florin0x01:
Based on cmake targetver definitions
The variable `tagartifact `must be defined (is defined in the azure pipelines CI build) : 
https://github.com/stream-labs/obs-studio-node/blob/6130a9d449172f242ac6e91ff9d92e634263b3a0/azure-pipelines.yml#L28
 Context: client and server need to be on same version to avoid any mismatches

Update:
Also added JS exception to display for `host` method called from frontend

### DIFF
--- a/obs-studio-client/source/controller.hpp
+++ b/obs-studio-client/source/controller.hpp
@@ -57,5 +57,5 @@ class Controller
 	private:
 	bool                         m_isServer = false;
 	std::shared_ptr<ipc::client> m_connection;
-	ProcessInfo                  procId;
+	ipc::ProcessInfo                  procId;
 };
--- a/obs-studio-client/source/controller.hpp
+++ b/obs-studio-client/source/controller.hpp
@@ -18,6 +18,7 @@
 
 #pragma once
 #include <memory>
+#include <map>
 #include <string>
 #include "ipc-client.hpp"
 #include <napi.h>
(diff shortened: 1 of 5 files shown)


florin0x01:
For SLOBS, the flow is as follows:

```obs.IPC.host(`slobs-${uuid()}`);  ```-> here the client spawns the server sending its version via commandline arg
   OSN server then checks against its own version and throws exception if mismatch
    Initialize OBS API -> the ```SLOBS_VERSION``` is sent to the crashhandler in order to have the correct GUI tag in sentry etc
    So I believe there is nothing to be done on SLOBS side?

``` 
const apiResult = obs.NodeObs.OBS_API_initAPI(
      'en-US',
      appService.appDataDirectory,
      electron.remote.process.env.SLOBS_VERSION,
    );
```

**Review by EddyGharbi — APPROVED** [review]

File: obs-studio-client/source/controller.hpp [hunk 1]
```diff
@@ -33,17 +34,38 @@ struct ProcessInfo
 {
 	uint64_t handle;
 	uint64_t id;
+	DWORD    exit_code;
 
-	ProcessInfo()
+	enum ExitCode
```
florin0x01: I think we can add this to lib-streamlabs-ipc , as it's more generic, but did not know where to put it exactly, so I left it here for now.

obs-studio-client/source/controller.hpp
```diff
@@ -33,17 +34,38 @@ struct ProcessInfo
...
-	}
+
+	typedef std::map<ProcessInfo::ExitCode, std::string> ProcessDescriptionMap;
+
+	private:
+		static ProcessDescriptionMap descriptions;
+		static ProcessDescriptionMap initDescriptions();
+
+	public:
+		ProcessInfo() : handle(0), id(0), exit_code(0)
```
florin0x01: same here, for ProcessInfo, maybe move to lib-streamlabs-ipc?

obs-studio-client/source/controller.hpp
EddyGharbi: Agree, seems better to have them in lib-streamlabs-ipc especially given the fact we are returning these specifics error codes from the server below. I would put them in ipc.h

obs-studio-server/source/main.cpp
```diff
@@ -192,18 +211,13 @@ int main(int argc, char* argv[])
...
-		std::string socketPath = "";
-#ifdef __APPLE__
-		socketPath = "/tmp/";
-#endif
-		socketPath += argv[1];
 		myServer.initialize(socketPath.c_str());
 	} catch (std::exception& e) {
 		std::cerr << "Initialization failed with error " << e.what() << "." << std::endl;
-		return -2;
+		return 253;
```
EddyGharbi: See above, I wouldn't put these like this but I would put them in constant variables in lib-streamlabs-ipc

florin0x01:
@EddyGharbi added JS check and some other changes.

PR #805, opened 2020-12-11
Handshake versioncontrol applefix

florin0x01:
Fixes Mac start of slobs. (wrong number of params)


obs-studio-server/source/main.cpp
```diff
@@ -142,10 +143,31 @@ int main(int argc, char* argv[])
 	g_util_osx = new UtilInt();
 	g_util_osx->init();
 #endif
+	std::string socketPath      = "";
```
florin0x01, line 146: Here

florin0x01:
Edit: Updated in https://github.com/stream-labs/obs-studio-node/pull/805/files so this should no longer be needed.

PR #846, opened 2021-02-02
Only call volmeters JS cbs if the source is not muted

EddyGharbi:


obs-studio-client/source/callback-manager.cpp
```diff
@@ -184,18 +184,21 @@ void globalCallback::worker()
 			for (auto vol: volmeters) {
 				VolmeterData* data     = new VolmeterData{{}, {}, {}};
 				size_t channels = response[index++].value_union.i32;
+				bool isMuted = response[index++].value_union.i32;
```
florin0x01, line 187: could be worthwhile to check the index if it's covered or not in the `response`

EddyGharbi: Unless I am misunderstanding what you mean, we don't have a good way to do this.

PR #847, opened 2021-02-03
Get devices api

EddyGharbi:


obs-studio-client/source/nodeobs_settings.cpp
```diff
@@ -447,6 +447,69 @@ Napi::Value settings::OBS_settings_getListCategories(const Napi::CallbackInfo& i
 	return categories;
 }
 
+Napi::Array devices_to_js(const Napi::CallbackInfo& info, std::vector<ipc::value> response)
+{
+	Napi::Array devices = Napi::Array::New(info.Env());
+
+	uint32_t js_array_index = 0;
+	uint64_t items = response[1].value_union.ui64;
```
florin0x01, line 455: maybe it's better to send `response` via const ref?

obs-studio-server/source/nodeobs_settings.cpp
```diff
@@ -3984,3 +3990,109 @@ void OBS_settings::saveGenericSettings(std::vector<SubCategory> genericSettings,
...
+	else
+		rval.push_back(ipc::value((uint64_t)items));
+
+	for (size_t idx = 0; idx < items; idx++) {
+		const char* description = obs_property_list_item_name(prop, idx);
+		const char* device_id = obs_property_list_item_string(prop, idx);
+
+		if (!description || !strcmp(description, "") ||
+			!device_id || !strcmp(device_id, "")) {
+			rval[1].value_union.ui64--;
```
florin0x01, line 4028: could this be `rval[0]` in case `rval.size()` is 1 ?

EddyGharbi: Our IPC format requires the first value `rval[0]` to be the error code. That's why I put the size the value after that.

obs-studio-client/source/nodeobs_settings.cpp
EddyGharbi: good point / good catch, making the change now.

PR #854, opened 2021-02-11
Releases all scenes, sources and scene items during a forced shutdown

EddyGharbi:


obs-studio-server/source/nodeobs_api.cpp
```diff
@@ -1407,6 +1407,32 @@ void OBS_API::destroyOBS_API(void)
...
+			obs_set_output_source(i, nullptr);
+
+		std::vector<obs_source_t*> sources;
+		osn::Source::Manager::GetInstance().for_each([&sources](obs_source_t* source)
+		{
+			sources.push_back(source);
+		});
+
+		for (const auto &source: sources) {
+			if (!strcmp(obs_source_get_id(source), "scene")) {
```
florin0x01, line 1421: check if source is null as well?

florin0x01 (approved):

PR #867, opened 2021-02-25
Add media control functionality

berkon:
Media file control functionality is already available in OBS library. I've added the corresponding bindings to the `osn::Input` class to make it available in Node.

**Usage:**
```
let input = osn.InputFactory.create ( 'ffmpeg_source', 'my-source', { local_file = 'my-video.mp4' } )

let duration = input.getDuration()  // get duration of media file im milliseconds
input.seek = 10000    // set play position to 10 sec
let pos = input.seek  // get play position in milliseconds
input.play()     // play media
input.pause()    // pause media
input.restart()  // restart media when ended
input.stop()     // stop media
```

### DIFF
--- a/obs-studio-client/source/input.cpp
+++ b/obs-studio-client/source/input.cpp
@@ -61,6 +61,7 @@ Napi::Object osn::Input::Init(Napi::Env env, Napi::Object exports) {
 			InstanceAccessor("deinterlaceFieldOrder", &osn::Input::GetDeinterlaceFieldOrder, &osn::Input::SetDeinterlaceFieldOrder),
 			InstanceAccessor("deinterlaceMode", &osn::Input::GetDeinterlaceMode, &osn::Input::SetDeinterlaceMode),
 			InstanceAccessor("filters", &osn::Input::Filters, nullptr),
+			InstanceAccessor("seek", &osn::Input::GetTime, &osn::Input::SetTime),
 
 			InstanceAccessor("configurable", &osn::Input::CallIsConfigurable, nullptr),
 			InstanceAccessor("properties", &osn::Input::CallGetProperties, nullptr),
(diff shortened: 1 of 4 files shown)


EddyGharbi:
Thank you for your contribution, this is very much appreciated! We were in the middle of a big release and I couldn't get to it sooner, sorry about that.

berkon:
No problem @EddyGharbi! I'm happy to hear that it is helpful! :-)

MizunoSakura:
So does this add a pause/play function to Streamlabs OBS or...?

berkon:
> So does this add a pause/play function to Streamlabs OBS or...?

My modification only adds the functionality to the library, but I don't know if the latest version of Streamlabs OBS already makes use of it.

MizunoSakura:
> > So does this add a pause/play function to Streamlabs OBS or...?
> 
> My modification only adds the functionality to the library, but I don't know if the latest version of Streamlabs OBS already makes use of it.

I'm... not exactly sure what the library is, but I can say Streamlabs OBS still doesn't have it. It's the one reason I have to use Gamecaster for certain things which is frustrating as it's not nearly as flexible with capturing things.

berkon:
@MizunoSakura, "libobs" is one of the main libraries of the [OBS](https://obsproject.com) project. **Streamlabs OBS** is different from **OBS** in that it is based on the [Electron ](https://www.electronjs.org) framework and thus programmed in Javascript. In order to be able to use the libobs functionality from the OBS project in Streamlabs OBS as well, a wrapper library is required which makes the C++ functions available to Javascript. This is exactly what this repo "obs-studio-node" is for.

**Long story short:** libobs had the media controls functionality already (and it can be used in OBS), Now I added the bindings to this wrapper library as well. But it is now up to Streamlabs when/if they will also put the media controls into the UI of Streamlabs OBS.

If you need the media controls, why don't you use OBS (I mean not Streamlabs OBS)? OBS has the same (and some more) functionality and usage is very similar to Streamlabs OBS.

MizunoSakura:
Mainly due to all the Streamlabs widgets which are kind of a pain to set up manually ^^; I may look into just downloading regular OBS too though then...

rosendolu:
hi @berkon , i am trying to  implement a  `video player` which needs to be set many videos `video1,video2,video3`  and start play in order. But seems no `end` event or `currentTime`  provided, any other way achieve that

berkon:
@rosendolu, its quite a while that I worked on this topic. But from what you say/ask for, I would say that you could get the length of a video with `input.getDuration()` and then periodically (maybe every second) check with a timer the progress. You can get the current play position with `input.seek`. See my description at the beginning of this thread.

rosendolu:
> @rosendolu, its quite a while that I worked on this topic. But from what you say/ask for, I would say that you could get the length of a video with `input.getDuration()` and then periodically (maybe every second) check with a timer the progress. You can get the current play position with `input.seek`. See my description at the beginning of this thread.

Thank you berkon, i tried this approach. `seek` step to  `40` when the video play end.Can i just use this value as the `toggle next video` trigger. I can't found any clue  about `40` from your code  .By the way `input.play()` not work when video play end, but `input.restart` start from 0, how can i get the `paused` or `playing` state


img_v3_02al_153de08a-2fd2-4eda-b7f6-cc99198e1fcg

PR #901, opened 2021-04-09
Set crash handler to save a memory dump on crash

summeroff:


obs-studio-server/source/nodeobs_api.cpp
```diff
@@ -562,6 +562,21 @@ std::vector<char> registerProcess(void)
 	return buffer;
 }
 
+std::vector<char> registerMemoryDump(void)
+{
+	std::vector<char> buffer;
+	uint8_t action     = 2;
```
EddyGharbi: Should we but this in a constant / enum, why did we use the value 2?

obs-studio-server/source/util-crashmanager.cpp
```diff
@@ -238,10 +241,63 @@ nlohmann::json RequestProcessList()
...
+
+	PSECURITY_DESCRIPTOR securityDescriptor = (PSECURITY_DESCRIPTOR) LocalAlloc(LPTR, SECURITY_DESCRIPTOR_MIN_LENGTH);
+	if (InitializeSecurityDescriptor(securityDescriptor, SECURITY_DESCRIPTOR_REVISION)) {
+		if (SetSecurityDescriptorDacl( securityDescriptor, TRUE, NULL, FALSE)) {
+			SECURITY_ATTRIBUTES eventSecurityAttr = {0};
+			eventSecurityAttr.nLength = sizeof(SECURITY_ATTRIBUTES);
+			eventSecurityAttr.lpSecurityDescriptor = securityDescriptor;
+			eventSecurityAttr.bInheritHandle = FALSE;
+
+			memoryDumpEvent = CreateEvent( &eventSecurityAttr, TRUE, FALSE, L"Global\\OBSMEMORYDUMPEVENT");
```
EddyGharbi: Might be better to call it `SLOBSMEMORYDUMPEVENT` so there is no confusion.

obs-studio-server/source/nodeobs_api.cpp
summeroff: before that we use only 0 and 1 as register and unregister commands. 
yes using enum make sense now even more. will update .

summeroff: done

obs-studio-server/source/util-crashmanager.cpp
summeroff: done.

EddyGharbi (approved):

PR #1013, opened 2021-10-05
Log long ipc

summeroff:


obs-studio-client/source/utility.cpp
```diff
@@ -129,7 +136,20 @@ void write_app_state_data(std::string app_state_path, std::string updated_status
 	}
 }
 
-void ipc_freez_callback(bool freez_detected, std::string app_state_path)
+void truncate_long_log_file(std::string log_file, size_t limit)
```
EddyGharbi: Am I misunderstanding or this function doesn't truncate but remove the file if the size is higher than the limit? A truncate function would only remove the excess no?

summeroff: I will update function.

summeroff: done.

EddyGharbi (approved):

PR #1016, opened 2021-10-07
Tag sentry report with name .dmp in s3

stevensoftware52:
This commit is dependent on a corresponding update to crash-handler-process
https://github.com/stream-labs/crash-handler/tree/sentry-memorydmps

- Add event SLOBSMEMORYDUMPEVENTFAIL
- Add event SLOBSMEMORYDUMPEVENTSUCCESS
- SignalMemoryDump returns true/false depending on successful upload to aws
- Rename functions/events for "starting" the upload prompt with "_Start"
- Filename is now decided here and sent to crash-handler-process for synchronization

### DIFF
--- a/obs-studio-server/source/util-crashmanager.cpp
+++ b/obs-studio-server/source/util-crashmanager.cpp
@@ -32,6 +32,7 @@
 #include <thread>
 #include <vector>
 #include <filesystem>
+#include <random>
 
 #ifdef WIN32
 #include "StackWalker.h"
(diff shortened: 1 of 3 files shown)


summeroff (approved):

obs-studio-server/source/util-crashmanager.cpp
```diff
@@ -600,6 +673,9 @@ void util::CrashManager::HandleCrash(std::string _crashInfo, bool callAbort) noe
 
 void util::CrashManager::SetReportServerUrl(std::string url)
 {
+	// dev environment
+	//url = "https://o114354.ingest.sentry.io/api/252950/minidump/?sentry_key=8f444a81edd446b69ce75421d5e91d4d";
```
EddyGharbi, line 677: What is this? Is commenting out intended?

EddyGharbi (approved):

obs-studio-server/source/util-crashmanager.cpp
stevensoftware52: It's the development DSN
https://github.com/stream-labs/streamlabs-obs/blob/fa386be6d91cb3708c460d034f3fc0ff47197dfb/app/app.ts#L45

I left it as a comment for convenience when/if debugging

PR #1025, opened 2021-10-22
apply try catch to more js callbacks

summeroff:


obs-studio-client/source/callback-manager.cpp
```diff
@@ -99,6 +99,7 @@ void globalCallback::worker()
 	auto sources_callback = []( Napi::Env env, 
 			Napi::Function jsCallback,
 			SourceSizeInfoData* data ) {
+		try {
 		Napi::Array result = Napi::Array::New(env, data->items.size());
```
EddyGharbi: Style issue, missing spaces due to the new block.

summeroff: updated in other places too

EddyGharbi (approved):

PR #1100, opened 2022-04-04
Osn video

EddyGharbi:
### Description
Expose a C++ API for the video context libobs API. The old API will remain until the client fully transitions. For the time being `Video.set()` will also save the video settings in the `basic.ini` config file, this is done in order for the old and new API to still working together.

### Motivation and Context
Nodeobs settings is hard to use and we want to transition into having these settings 100% from the client.

### How Has This Been Tested?
I created automated tests using this new AP, all tests are passing.

### Types of changes

### Checklist:
- [X] The code has been tested.
- [X] All commit messages are properly formatted and commits squashed where appropriate.


obs-studio-server/source/osn-video.cpp
```diff
@@ -52,4 +64,199 @@ void osn::Video::GetTotalFrames(
...
+    video.base_width = baseWidth;
+    video.base_height = baseHeight;
+    video.output_width = outputWidth;
+    video.output_height = outputHeight;
+    video.output_format = (video_format)outputFormat;
+    video.colorspace = (video_colorspace)colorspace;
+    video.range = (video_range_type)range;
+    video.scale_type = (obs_scale_type)scaleType;
+    video.adapter = 0;
+	video.gpu_conversion = true;
```
summeroff: only line with tab instead of space at the beginning.

obs-studio-client/source/video.cpp
```diff
@@ -76,3 +78,65 @@ Napi::Value osn::Video::encodedFrames(const Napi::CallbackInfo& info)
...
+	video.Set("range", response[9].value_union.ui32);
+	video.Set("scaleType", response[10].value_union.ui32);
+
+	return video;
+}
+
+void osn::Video::set(const Napi::CallbackInfo& info, const Napi::Value &value)
+{
+	Napi::Object video = value.ToObject();
+	if (!video || !video.IsObject()) {
```
summeroff: IsObject() checked after ToObject()

EddyGharbi: Good catch, fixed

obs-studio-server/source/osn-video.cpp
EddyGharbi: Good catch, fixed

summeroff (approved):

PR #1120, opened 2022-05-03
Add function 'slowUncachedSettings'

stevensoftware52:
Allows frontend to skip cache and ask for a copy of the latest settings in backend memory

### Description
Allows frontend to skip cache and ask for a copy of the latest settings in backend memory

### Motivation and Context
The frontend for the join plugin needs to be able to get the exact latest settings because the exchange of adding/removing settings is the communication structure we're planning to use.

### How Has This Been Tested?
I replaced all frontend .ts instances of ().settings with ().slowUncachedSettings(), attached to electron and waited for my new function to be hit. The return value is a json blob returned from ipc so this should be the latest settings.

### Types of changes
Add slowUncachedSettings(): ISettings; to the export interface ISource
Add the function to respective client cpp files, and reused the server functionality of "GetSettings"
_NOTE: server changes weren't required because the caching logic appears to be at the client via **SourceDataInfo::settingsChanged** - The server function GetSettings just returns **obs_source_get_settings**, so I've re-used it._

### Checklist:

### DIFF
--- a/js/module.ts
+++ b/js/module.ts
@@ -1190,7 +1190,7 @@ export interface ISource extends IConfigurable, IReleasable {
      * Function to get latest version of settings
 	 * Expensive, shouldn't be used unless sure
      */
-    slowUncachedSettings(): ISettings;
+    readonly slowUncachedSettings: ISettings;
 }
 
 export interface IFaderFactory {
--- a/js/module.ts
+++ b/js/module.ts
@@ -1185,6 +1185,12 @@ export interface ISource extends IConfigurable, IReleasable {
      * Easy way to disable a filter.
      */
     enabled: boolean;
+	
+    /**
+     * Function to get latest version of settings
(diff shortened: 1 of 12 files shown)


summeroff (approved):

js/module.ts
```diff
@@ -1185,6 +1185,12 @@ export interface ISource extends IConfigurable, IReleasable {
      * Easy way to disable a filter.
      */
     enabled: boolean;
+	
+    /**
+     * Function to get latest version of settings
+	 * Expensive, shouldn't be used unless sure
+     */
+    slowUncachedSettings(): ISettings;
```
avacreeth: I notice this is a function - should it be a property like the existing `settings` property?

i.e. it would be `readonly slowUncachedSettings: ISettings;`

EddyGharbi (approved):

PR #1123, opened 2022-05-11
Callhandlerfunction

stevensoftware52:
### Description
Adds a frontend function for obs_source's `callHandler` which accepts two parameters: the name of a proc_handler function to execute, and parameters to send into that function. 

### Motivation and Context
This allows plugins to register functions with a custom purpose that the frontend can then execute at its own leisure. Specifically this is required for the Join plugin.

### How Has This Been Tested?
1. Modified a filter plugin to register 'func_send_transport_response' via proc_handler
2. Added to `source-filters.ts` after filter creation `obsFilter.callHandler('func_send_transport_response', 'params');`
3. Breakpoint in plugin function was hit, string 'params' was observed in plugin, then my assigned return value string "hello world" was seen at osn client breakpoint for returned napi object

### Types of changes
1. OSN Client/Server 
2. module.ts `interface ISource`

### Checklist:

### DIFF
--- a/obs-studio-client/source/isource.cpp
+++ b/obs-studio-client/source/isource.cpp
@@ -486,7 +486,7 @@ Napi::Object osn::ISource::CallHandler(const Napi::CallbackInfo& info, uint64_t 
 
 	if (!conn)
 	{
-		result.Set("error", Napi::Boolean::New(info.Env(), true));
+		Napi::TypeError::New(info.Env(), "IPC Connection Failed").ThrowAsJavaScriptException();
 		return result;
 	}
 
@@ -495,7 +495,7 @@ Napi::Object osn::ISource::CallHandler(const Napi::CallbackInfo& info, uint64_t 
 
 	if (!ValidateResponse(info, response))
 	{
-		result.Set("error", Napi::Boolean::New(info.Env(), true));
+		Napi::TypeError::New(info.Env(), "Invalid IPC Response").ThrowAsJavaScriptException();
 		return result;
 	}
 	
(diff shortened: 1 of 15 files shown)


obs-studio-client/source/isource.cpp
```diff
@@ -475,6 +475,33 @@ void osn::ISource::SetMuted(const Napi::CallbackInfo& info, const Napi::Value &v
...
+		result.Set("error", Napi::Boolean::New(info.Env(), true));
+		return result;
+	}
+
+	std::vector<ipc::value> response =
+	    conn->call_synchronous_helper("Source", "CallHandler", { ipc::value(id), ipc::value(fuction_name), ipc::value(fuction_input) });
+
+	if (!ValidateResponse(info, response))
+	{
+		result.Set("error", Napi::Boolean::New(info.Env(), true));
```
EddyGharbi: Is there something specific the client needs to do in case of an error? In this case, throwing a JS exception might be best.

obs-studio-client/source/scene.cpp
```diff
@@ -74,6 +74,7 @@ Napi::Object osn::Scene::Init(Napi::Env env, Napi::Object exports) {
 			InstanceMethod("sendMouseWheel", &osn::Scene::CallSendMouseWheel),
 			InstanceMethod("sendFocus", &osn::Scene::CallSendFocus),
 			InstanceMethod("sendKeyClick", &osn::Scene::CallSendKeyClick),
+			InstanceMethod("call_handler", &osn::Scene::CallCallHandler),
```
summeroff: a name of new method in a different style from other methods. Is it intentional?

stevensoftware52: A mistake on my part - had the backend style in my mind while I was writing this.

summeroff (approved):

EddyGharbi (approved):

PR #1142, opened 2022-07-02
Node obs log performance 2

eugen15:
### Description
This change disables LOG_DEBUG messages for log files created in `slobs-client\node-obs\logs` for the release build. It does not touch any other log destination like `OBS_API::LogReport` and `OutputDebugStringW`.
You can still enable LOG_DEBUG by creating a file _slobs-client/**enable-debug-logs**_ even if you have compiled the application with the release config. See the full typical paths below:
* **Windows**: _C:\Documents and Settings\USER_NAME\AppData\Roaming\slobs-client\\**enable-debug-logs**_
* **macOS**: _/Users/USER_NAME/Library/Application Support/slobs-client/**enable-debug-logs**_

### Motivation and Context
By default, we should not write debug logs to user machines. They do not need it.

### How Has This Been Tested?
1. Start Streamlabs Desktop, do some actions, close it.
2. View newly created logs in the folder:
* **Windows**: C:\Documents and Settings\YOUR_USERNAME\AppData\Roaming\slobs-client\node-obs\logs
* **macOS**: /Users/YOUR_USERNAME/Library/Application Support/slobs-client/node-obs/logs
3. Check that the file lines do not have `[Debug]`. They should have something like `[INFO]`, `[ERROR]`, etc. The old files may still have `[Debug]`.
4. Create the **enable-debug-logs** file in the appropriate folder (see examples above).
5. Start Streamlabs Desktop again, do some actions, close it.
6. Now you may see the `[Debug]` in the newly created log files.

### Types of changes
- New feature (it only affects developers)

### Checklist:


obs-studio-server/source/nodeobs_api.cpp
```diff
@@ -403,7 +409,7 @@ std::chrono::high_resolution_clock::time_point tp = std::chrono::high_resolution
 static void                                    node_obs_log(int log_level, const char* msg, va_list args, void* param)
 {
 	if (param == nullptr)
-		return;
+		return;
```
EddyGharbi: Extra space

EddyGharbi:
Instead of checking if a file exists, wouldn't it be easier for developers if we were using an environment variable? I reset my cache a lot and I will have to think every time to create this file.

obs-studio-server/source/nodeobs_api.cpp
```diff
@@ -717,6 +727,34 @@ void writeCrashHandler(std::vector<char> buffer)
...
+	// https://stackoverflow.com/questions/135688/setting-environment-variables-on-os-x
+	char* envValue = getenv("SL_DESKTOP_ENABLE_DEBUG_LOGS");
+	if (envValue != nullptr) {
+		if (astrcmpi(envValue, "on") == 0 || astrcmpi(envValue, "yes") == 0) {
+			return true;
+		}
+	}
+	// Even if the environment variable is set and the logs are turned off,
+	// we still check for the file to avoid issues with finding where the variable is defined.
+	const std::string filename = appdata + "/enable-debug-logs";
```
EddyGharbi, line 751: Why do the debug logs need to go to a separate file when it's enabled? Could we not still use the same log file?

eugen15: This is not a log file. This is a file to check if debug logs are enable. I would like to have both ways (the environment variable and the file) because most likely on macOS the environment variable way is not convenient.

obs-studio-server/source/nodeobs_api.cpp
eugen15: Fixed

EddyGharbi (approved):

PR #1145, opened 2022-07-07
Rework device api

EddyGharbi:
### Description
Change to use Windows native API to enumerate available DSHOW devices.

### Motivation and Context
The method of creating dummy devices to then enumerate through the properties (1) doesn't work well (2) create some issues on the side and (3) not good in terms of performance. We are effectively duplicating the device enum logic from OBS but that's ok since it's very isolate and relatively small.

### How Has This Been Tested?
With QA and from the client to make sure we are correctly enumerating all the devices.

### Types of changes
Performance enhancement (non-breaking change which improves efficiency)

### Checklist:
- [X] The code has been tested.


obs-studio-server/source/nodeobs_settings.cpp
```diff
@@ -4310,17 +4316,145 @@ void getDevices(const char *source_id, const char *property_name, std::vector<ip
...
+		blog(LOG_ERROR, "Failed to get device count");
+
+	rval[1] = ipc::value((uint32_t)count);
+
+	for (UINT i = 0; i < count; i++) {
+		ComPtr<IMMDevice> device;
+		CoTaskMemPtr<WCHAR> w_id;
+
+		res = collection->Item(i, device.Assign());
+		if (FAILED(res))
```
summeroff: if we skip some devices. "count" put inside "rval[1]" became incorrect.

EddyGharbi: oops, good catch, you're right. Fixing

EddyGharbi: fixed

summeroff (approved):

summeroff (approved):

PR #1166, opened 2022-08-23
Add scene item rotation handle and a request to enable drawing it

eugen15:
### Description
- Added a rotation handle for scene items
- Added a client/server request to enable/disable drawing the rotation handle

### Motivation and Context
It prepares the backend for custom scene item rotation.

### How Has This Been Tested?
- Tested with different source item rotation angles to be sure the handle is shown correctly.
- Tested with custom desktop code to test the client/server request works correctly.

### Types of changes

### Checklist:

### DIFF
--- a/obs-studio-server/source/nodeobs_common.cpp
+++ b/obs-studio-server/source/nodeobs_common.cpp
@@ -222,6 +222,11 @@ void OBS_content::Register(ipc::server& srv)
 	    OBS_content_setDrawGuideLines));
 
 	cls->register_function(std::make_shared<ipc::function>(
+	    "OBS_content_setDrawRotationHandle",
+	    std::vector<ipc::type>{ipc::type::String, ipc::type::Int32},
+	    OBS_content_setDrawRotationHandle));
+
+	cls->register_function(std::make_shared<ipc::function>(
 	    "OBS_content_createIOSurface",
 	    std::vector<ipc::type>{ipc::type::String},
 	    OBS_content_createIOSurface));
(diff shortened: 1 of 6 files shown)


obs-studio-server/source/nodeobs_display.cpp
```diff
@@ -305,6 +308,52 @@ OBS::Display::Display()
 	*v.color = 0xFFFFFFFF;
 	m_boxTris->Update();
 
+	// Rotation handle line
+	m_rotHandleLine = std::make_unique<GS::VertexBuffer>(5);
+	m_rotHandleLine->Resize(5);
+	v = m_rotHandleLine->At(0);
+	vec3_set(v.position, 0.5f - 0.34f / HANDLE_RADIUS, 0.5f, 0);
```
summeroff, line 315: "0.5f - 0.34f / HANDLE_RADIUS"
compiler will optimize it anyway but moving to a const variable will make code cleaner

eugen15: You are right! I will fix it.

eugen15: Actually, it turned out it is more complicated. I cannot convert `0.5f - 0.34f` to a constant variable because `0.34f / HANDLE_RADIUS` goes first (division goes first). If I pre-calculate the numbers entirely and remove `HANDLE_RADIUS` at all, then I have to do the same in `OBS::Display::DrawRotationHandle` because it depends on `m_rotHandleLine` and `HANDLE_RADIUS`.  Then, in some time, it will be difficult to understand why we use the numbers. OR do you mean something else?

summeroff: if it will complicate code then it is good to keep as is.

summeroff (approved):

obs-studio-server/source/nodeobs_common.cpp
```diff
@@ -654,6 +659,24 @@ void OBS_content::OBS_content_setDrawGuideLines(
...
+    std::vector<ipc::value>&       rval)
+{
+	// Find Display
+	auto it = displays.find(args[0].value_str);
+	if (it == displays.end()) {
+		rval.push_back(ipc::value((uint64_t)ErrorCode::Error));
+		rval.push_back(ipc::value("Display key is not valid!"));
+		return;
+	}
+	it->second->SetDrawRotationHandle((bool)args[1].value_union.i32);
```
EddyGharbi: Should be `value_union.ui32`

eugen15: Fixed

PR #1299, opened 2023-05-26
Cannot disrupt video ptr inside obs while outputs are connecting

stevensoftware52:
### Description
The pointer to obs->video cannot be disrupted during the rtmp-stream connection phase.

### Motivation and Context
Changing video settings will trigger the video object to be recreated. If rtmp is in the middle of connecting, it will crash. This prevents that.

### How Has This Been Tested?
See associated PR: https://github.com/stream-labs/obs-studio/pull/517

### Types of changes

### Checklist:

### DIFF
--- a/obs-studio-server/source/nodeobs_service.cpp
+++ b/obs-studio-server/source/nodeobs_service.cpp
@@ -459,8 +459,19 @@ int OBS_service::resetVideoContext(bool reload, bool retryWithDefaultConf)
 	return errorcode;
 }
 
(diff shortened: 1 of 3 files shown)


summeroff:
Sould osn::Video::SetVideoContext do the same as osn::Video::RemoveVideoContext. 
In this requests probably better to return error instead force stopping. User might want to change video settings but did not think about his streaming status.

summeroff (approved):

PR #1328, opened 2023-07-19
Refined thread-safety for MemoryManager

avoitenko-logitech:
### Description
- Refined thread-safety model for MemoryManager.
- Fixed multiple data races + 1 undefined behavior.
- Core logic and algorithms left unchanged, but work with locks and threads was almost completely rewritten.
- Additionally made some modern C++ code improvements
- Removed watcher as it was unused.

### Motivation and Context
An attempt to fix crashes.
I'll leave comments with descriptions of problems of the old code.

### How Has This Been Tested?
Unit tests + logs analysis

### Types of changes

### DIFF
--- a/obs-studio-server/source/memory-manager.cpp
+++ b/obs-studio-server/source/memory-manager.cpp
@@ -19,15 +19,31 @@
 #include "memory-manager.h"
 #include "nodeobs_api.h"
 
+struct MemoryManager::source_info {
+	bool cached = false;
+	uint64_t size = 0;
+	obs_source_t *source = nullptr;
+	std::vector<std::thread> workers;
+	std::mutex mtx;
+	bool have_video = false;
+};
+
+MemoryManager &MemoryManager::GetInstance()
+{
+	static MemoryManager instance;
+	return instance;
+}
+
 MemoryManager::MemoryManager()
 {
+	blog(LOG_INFO, "MemoryManager: constructor called");
 #ifdef WIN32
 	MEMORYSTATUSEX statex;
 	statex.dwLength = sizeof(statex);
 
-	if (GlobalMemoryStatusEx(&statex)) {
+	if (::GlobalMemoryStatusEx(&statex)) {
 		available_memory = statex.ullTotalPhys;
-		allowed_cached_size = std::min((uint64_t)LIMIT, (uint64_t)available_memory / 2);
+		allowed_cached_size = std::min<uint64_t>(LIMIT, available_memory / 2);
 	} else {
 		available_memory = 0;
 		allowed_cached_size = LIMIT;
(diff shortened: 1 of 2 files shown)


obs-studio-server/source/memory-manager.cpp
```diff
@@ -148,218 +155,171 @@ void MemoryManager::addCachedMemory(source_info *si)
...
+	si.cached = false;
 
-	updateSource(si->source, false);
+	updateSource(si.source, false);
 
 	if (!cacheNewFiles || current_cached_size >= allowed_cached_size)
 		return;
 
-	for (auto data : sources) {
-		if (strcmp(obs_source_get_name(si->source), data.first) != 0 && shouldCacheSource(data.second))
```
avoitenko-logitech, line 174: [OLD CODE]
Data race on `shouldCacheSource` because in does not lock mutex inside.
And mutex acquired above `std::unique_lock<std::mutex> ulock(si->mtx);` in the old code is the different one and not related to current `data` object.

obs-studio-server/source/memory-manager.cpp
```diff
@@ -148,218 +155,171 @@ void MemoryManager::addCachedMemory(source_info *si)
...
 		if (worker.joinable())
 			worker.join();
 	}
 
-	mtx.lock();
-
-	it->second->workers.clear();
-	removeCachedMemory(it->second, true);
-
-	free(it->second);
```
avoitenko-logitech, line 306: [OLD CODE]
Undefined behavior which results in memory leak.
Object was allocated with C++ `new` operator and freed with  C `free` => no destructors call for classes like `std::vector` to free memory.

obs-studio-server/source/memory-manager.cpp
```diff
@@ -148,218 +155,171 @@ void MemoryManager::addCachedMemory(source_info *si)
...
-		return;
-	const char *source_id = obs_source_get_id(source);
-	if (!source_id)
-		return;
-	if (strcmp(source_id, "ffmpeg_source") != 0)
+	if (!isSourceValid(source)) {
 		return;
+	}
 
-	mtx.lock();
```
avoitenko-logitech, line 286: [OLD CODE]
Multiple data races here as we do not hold  `mtx` and `it->second.mtx` mutexes until deinit ends for usability reasons.
Few the most prominent:
- When someone is called `unregisterSource` it is possible to add new `worker` while we are waiting for threads to join => possibly corrupted `std::vector` of `workers` during iteration. This seems to be the reason of original crashes.
-  Call of `join` from multiple threads is undefined behavior (in reality it is a dead lock)
-  Guaranteed crash on dangling `it` if 2 different threads called `unregisterSource` at the same time for the same source and the second thread not hanged on `join` attempt because  worker threads were joined and cleaned at that point. It is possible BTW in real life, but chance is little.

obs-studio-server/source/memory-manager.cpp
```diff
@@ -148,218 +155,171 @@ void MemoryManager::addCachedMemory(source_info *si)
...
-		return;
+		if (!looping || !local_file) {
+			return;
+		}
+
+		need_get_size = si.size == 0;
 	}
-	si->mtx.unlock();
 
-	if (si->size == 0) {
```
avoitenko-logitech, line 194: [OLD CODE]
Data race on `si` item because `si->mtx.unlock();` was called above.

obs-studio-server/source/memory-manager.cpp
```diff
@@ -148,218 +155,171 @@ void MemoryManager::addCachedMemory(source_info *si)
...
-		uint32_t retry = MAX_POOLS;
+	if (need_get_size) {
+		uint32_t retry = MAX_POLLS;
 		while (retry > 0) {
-			mtx.lock();
```
avoitenko-logitech, line 197: [OLD CODE]
Unnecessarily locked `mtx` mutex here. `calculateRawSize` does not use data from `this` and it is guaranteed that during deletion all client threads should be joined before cleaning up, so no dangling object reference here during removal.

obs-studio-server/source/memory-manager.cpp
```diff
@@ -148,218 +155,171 @@ void MemoryManager::addCachedMemory(source_info *si)
...
-			si->mtx.unlock();
-			mtx.unlock();
+			si.mtx.unlock();
 
 			retry--;
 			std::this_thread::sleep_for(std::chrono::milliseconds(500));
 		}
 	}
-	std::unique_lock<std::mutex> ulock(mtx);
-	si->mtx.lock();
```
avoitenko-logitech, line 214: [OLD CODE]
Potential dead lock.
Disclaimer: it should not happen in the current code, but... as the code changes, some day different threads try to lock mutexes in different order and here we are.

summeroff (approved):

PR #1350, opened 2023-09-06
Notify crash handler about GPU reset exception

eugen15:
### Checklist:

- [ ] The code has been tested.
- [ ] All commit messages are properly formatted and commits squashed where appropriate.
- [ ] I have included updates to all appropriate documentation.


obs-studio-server/source/nodeobs_api.cpp
```diff
@@ -699,6 +707,34 @@ std::vector<char> crashedProcess(uint32_t crash_id)
...
+
+	// Pack	
+	uint32_t offset = 0;
+	memcpy(buffer.data(), &messageAction, sizeof(messageAction));
+	offset++;
+	memcpy(buffer.data() + offset, &messageCode, sizeof(messageCode));
+	offset += sizeof(messageCode);
+	memcpy(buffer.data() + offset, &messageCodeDescSize, sizeof(messageCodeDescSize));
+	offset += sizeof(messageCodeDescSize);
+	memcpy(buffer.data() + offset, messageCodeDesc.data(), messageCode);
```
avoitenko-logitech, line 731: Seems like a typo here. Last parameter should be size, but it is `messageCode`

avoitenko-logitech: Should be `messageCodeDescSize`

PR #1381, opened 2023-12-13
Handling unable to capture game via window case

avoitenko-logitech:
### Description
Modified game capture module to handle the case when it is not possible to capture a window because it is not a game

### Motivation and Context
Added support of info type feature of text properties
Using both placeholder and compatibility info to notify user.
To refresh UI trick with custom flag was used.

### How Has This Been Tested?
Manually

### Types of changes

### DIFF
--- a/obs-studio-client/source/properties.hpp
+++ b/obs-studio-client/source/properties.hpp
@@ -23,6 +23,7 @@
 #include <napi.h>
 #include <unordered_map>
 #include "utility-v8.hpp"
+#include "obs-property.hpp"
 
 namespace osn {
 struct Property {
@@ -73,13 +74,8 @@ struct NumberProperty : Property {
 };
 
 struct TextProperty : Property {
-	enum class Type {
-		DEFAULT,
-		PASSWORD,
-		MULTILINE,
-	};
-
-	Type field_type;
+	obs::TextProperty::TextType field_type;
+	obs::TextProperty::InfoType info_type;
 	std::string value;
 };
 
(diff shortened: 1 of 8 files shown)


obs-studio-client/source/properties.hpp
```diff
@@ -79,7 +79,14 @@ struct TextProperty : Property {
 		MULTILINE,
 	};
 
+	enum class InfoType {
```
summeroff: it will be duplicate of InfoType in source/obs-property.hpp

source/obs-property.hpp
```diff
@@ -127,7 +127,15 @@ struct TextProperty : Property {
 		Password,
 		MultiLine,
 	};
+
+	enum class InfoType : uint8_t {
```
summeroff, line 132: As values will be mapped to ETextInfoType better to put values here too explicitly.

avoitenko-logitech: Fixed

obs-studio-client/source/properties.hpp
avoitenko-logitech: I just followed existing approach here.
For the enum
```
	enum class Type {
		DEFAULT,
		PASSWORD,
		MULTILINE,
	};
```
there is a duplicate in `source/obs-property.hpp`
```
	enum class TextType : uint8_t {
		Default,
		Password,
		MultiLine,
	};
```

avoitenko-logitech: Should I avoid a code duplication?

summeroff: yes. if its possible better to not not have duplicated types.

avoitenko-logitech: Fixed

summeroff (approved):

summeroff (approved):

PR #1385, opened 2023-12-23
Apply settings back to created source after requesting properties from FE

summeroff:
### Description
Apply settings back to created source when properties get requested.

### Motivation and Context
Current implementation of utility::ProcessProperties does not change updateSource flag. 
It leads to missed settings update. 

Some values for properties gets populated only inside obs_source_properties. Like the list of displays. And without update source will use non working default values.  

### Types of changes

### Checklist:


avoitenko-logitech (approved):

rosendolu:
Hi @avoitenko-logitech @summeroff  is there a way to update `source settings` after created? I try to update settings fileds but not work. `sceneItem position` works well 
img_v3_02ak_d09dd497-c628-42c5-a09a-10acd52415dg

Even if called the update methods 
`sceneItem.source.update(sceneItem.source.settings)`

PR #1518, opened 2025-04-14
Implementing transition events

avoitenko-logitech:
### Description
- Implemented transition events (like start/stop) to be used in the client.
- Additionally fixed memory leak related to unprocessed source size events.

### Motivation and Context
This change is needed to better handle transitions-related logic and not rely on timers in client.

### How Has This Been Tested?
Manually, Windows only

### Types of changes

### DIFF
--- a/obs-studio-server/source/callback-manager.cpp
+++ b/obs-studio-server/source/callback-manager.cpp
@@ -26,8 +26,13 @@
 #include "osn-source.hpp"
 #include "osn-volmeter.hpp"
 
-std::mutex sources_sizes_mtx;
-std::map<std::string, SourceSizeInfo *> sources;
+#include <memory>
+
+std::mutex sources_mtx;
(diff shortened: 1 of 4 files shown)


obs-studio-server/source/callback-manager.cpp
```diff
@@ -65,10 +72,31 @@ void CallbackManager::GlobalQuery(void *data, const int64_t id, const std::vecto
...
+	rval.push_back(ipc::value((uint32_t)0));
+
+	if (!transitions.empty()) {
+		transitions_mtx.lock();
+		uint32_t size = 0;
+
+		for (auto &item : transitions) {
+			TransitionInfo *ti = item.second.get();
+
+			for (const auto &e : ti->events) {
```
sandboxcoder, line 89: (Trivial/Optional) Rename e -> `event`

sandboxcoder (approved):

summeroff (approved):

PR #1536, opened 2025-06-12
Add support for obs30 virtual cam

sandboxcoder:
### Description

* Add api function to create the new virtual cam obs-source
* Add new function to detect if the virtualcam system extension is installed locally
* Propagate signal to js indicating when the virtual cam is initialized
* See related slobs pull request [here](https://github.com/streamlabs/obs-studio/pull/649).

### Motivation and Context
Exposes the obs-30.2 virtual cam

### How Has This Been Tested?
* In OSN, added a small test case that initialized the virtual cam. However, this only works locally due to System extension(s) security restrictions (cant install those on a build agent).
* In Streamlabs-Desktop, verified this works in non-codesigned builds.
* Tested mostly outside of app bundles via the VIRTUALCAM_BYPASS_SYSTEM_CHECK environment var (I updated cmake to look for this var when its building the project for xcode). Also tested within an app bundle but w/o codesign.

I did not submit the local virtualcam test case (since it wont run on a build agent). The code looks like this:

```
    it('Test virtualcam output', async function() {
        if (!process.env.SLOBS_LOCAL_DEVELOPER || !obs.isDarwin()) {
            logInfo(testName, "Skipping this test it must be ran locally on a macOS Apple developer machine.")
            this.skip(); // This should only be ran locally because this test installs a SystemExtension which will not work on a VM
        }

        // Registers the global callback.
        // This step must be invoked before OBS_service_createVirtualCam
        obs.connectOutputSignals();

        // Special step on MacOS to create the output source.
        // Behind the scenes, this will install the SystemExtension
        osn.NodeObs.OBS_service_createVirtualCam();

        const signalInfo = await obs.getNextSignalInfo(
            EOBSOutputType.VirtualCam, EOBSOutputSignal.ReconnectSuccess);
        expect(signalInfo.type).to.equal(
            EOBSOutputType.VirtualCam, GetErrorMessage(ETestErrorMsg.VirtualCamStoppedWithError));

        // On MacOS, we should check this error string and display it to inform the user.
        let errorMsg = osn.NodeObs.OBS_service_startVirtualCam();
        expect(errorMsg === undefined);
        osn.NodeObs.OBS_service_stopVirtualCam();

        const installationStatus = osn.NodeObs.OBS_service_isVirtualCamPluginInstalled();
        expect(installationStatus).to.equal(2); // expect plugin is installed
    });
```
### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,7 +22,7 @@ env:
   SLGenerator: Visual Studio 17 2022
   SLDistributeDirectory: distribute
   SLFullDistributePath: "streamlabs-build.app/distribute" # The .app extension is required to run macOS tests correctly.
-  LibOBSVersion: 30.2.4sl21
+  LibOBSVersion: 30.2.4sl21-vcam5
   PACKAGE_NAME: osn
 
 jobs:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,7 +22,7 @@ env:
   SLGenerator: Visual Studio 17 2022
   SLDistributeDirectory: distribute
   SLFullDistributePath: "streamlabs-build.app/distribute" # The .app extension is required to run macOS tests correctly.
-  LibOBSVersion: 30.2.4sl22-vcam
+  LibOBSVersion: 30.2.4sl22-vcam1
   PACKAGE_NAME: osn
 
 jobs:
--- a/obs-studio-client/source/nodeobs_service.cpp
+++ b/obs-studio-client/source/nodeobs_service.cpp
@@ -36,10 +36,8 @@
 #include <shellapi.h>
 
 #define TOTALBYTES 8192
-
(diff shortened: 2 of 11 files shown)


.github/workflows/main.yml
```diff
@@ -22,7 +22,7 @@ env:
   SLGenerator: Visual Studio 17 2022
   SLDistributeDirectory: distribute
   SLFullDistributePath: "streamlabs-build.app/distribute" # The .app extension is required to run macOS tests correctly.
-  LibOBSVersion: 30.2.4sl21
+  LibOBSVersion: 30.2.4sl21-vcam5
```
avoitenko-logitech: Please do not forget to ask Vladimir to make a proper release of OBS version.

avoitenko-logitech (approved):

.github/workflows/main.yml
sandboxcoder: Thank you, I will make sure to do that

PR #1598, opened 2025-11-06
update OSN::Source::Release to check the ref count before deleting the source

sandboxcoder:
### Description
Update OSN::Source::Release to **release** the source (either decrement ref count or remove it entirely only if ref count equals 0). 

Now, we can 'release' an input source before OR after removing it from an output channel. This is because OBS typically checks the ref count on an source before using it.

The example code below demonstrates how the ref count changes. Now, we can safely release the input source whenever we desire. Previously, you needed to manually remove the source from the channel before invoking `release`.
```
    it('Set source to output channel and get it', function () {
        // Creating input source
        const input = osn.InputFactory.create(EOBSInputTypes.ImageSource, 'test_osn_global_source');

        const channel = 1;
        osn.Global.setOutputSource(channel, input); // Set ref count to 2

        const nullSource : ISource = null;
        osn.Global.setOutputSource(channel, nullSource); // decrement ref count on source (ref count = 1)
        input.release(); // decrement ref count on source (ref count = 0)
    });
```

### Motivation and Context
It dawned on me after merging my [last PR](https://github.com/streamlabs/obs-studio-node/pull/1597) that we might've had a very tiny chance of a race condition of sorts if there's any high level Desktop Js code that might invoke `release` BEFORE removing it from the output channel. With this change, there is no possibility of any sort of race condition as minor as it was (we often invoke release right before setting the output channel to NULL back-to-back but still its good to cover this hole imo)? Granted, looking at Desktop JS code I think we were fine because we just happened to remove the source from the channel before deleting it. So this change is simply nice-to-have I think as long as it doesnt cause any issues

Another motivation was I think the behavior is a lot more clear since OBS makes a clear distinction between `release` and `remove`. Now this code will be consistent with the OBS api definition.

### Checklist:

### DIFF
--- a/obs-studio-server/source/osn-source.cpp
+++ b/obs-studio-server/source/osn-source.cpp
@@ -228,7 +228,7 @@ void osn::Source::Release(void *data, const int64_t id, const std::vector<ipc::v
 
 		obs_source_release(src);
 	} else {
-        obs_source_release(src);
+		obs_source_release(src);
 	}
 
 	rval.push_back(ipc::value((uint64_t)ErrorCode::Ok));
--- a/obs-studio-server/source/osn-source.cpp
+++ b/obs-studio-server/source/osn-source.cpp
@@ -228,7 +228,7 @@ void osn::Source::Release(void *data, const int64_t id, const std::vector<ipc::v
 
 		obs_source_release(src);
 	} else {
-		obs_source_remove(src);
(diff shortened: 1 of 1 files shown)


sandboxcoder:
Unfortunately, this PR appears to undo this change: https://github.com/streamlabs/obs-studio-node/pull/1297.

PR #1619, opened 2026-01-14
settings: enable "rescale output" dropdown on Windows/macOS

sandboxcoder:
### Description
Enable "Rescale Output" dropdown on MacOS. Related [PR](https://github.com/streamlabs/obs-studio-node/pull/1617) where I enabled the checkbox but now we're revealing the dropdown as well.

image

### Motivation and Context
Restoring legacy feature from 1.19.6 where you could rescale output

### Types of changes

### Checklist:

### DIFF
--- a/obs-studio-server/source/nodeobs_settings.cpp
+++ b/obs-studio-server/source/nodeobs_settings.cpp
@@ -1892,12 +1892,7 @@ SubCategory OBS_settings::getAdvancedOutputStreamingSettings(config_t *config, b
 	rescale.currentValue.resize(sizeof(doRescale));
 	memcpy(rescale.currentValue.data(), &doRescale, sizeof(doRescale));
 	rescale.sizeOfCurrentValue = sizeof(doRescale);
-
-#if defined(WIN32)
-	rescale.visible = (encoderCurrentValue == ENCODER_NVENC_H264_TEX);
-#else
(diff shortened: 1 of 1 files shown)


aleksandr-voitenko (approved):

mhoyer-streamlabs (approved):
Approving but curious about Mac always being visible but only for a certain encoder on Windows, why is that?

sandboxcoder:
> Approving but curious about Mac always being visible but only for a certain encoder on Windows, why is that?

Honestly, I think the logic is inverted the more I think about it?

git blame says it got changed [here](https://github.com/streamlabs/obs-studio-node/commit/31a68f9da408531d492d9afbe9ad4a4ed60c3586#diff-b633b958b14d68adcc068ee16752b109045a243138c1b34c7d8044f655be2f2c).
image

Maybe what we want here is this? Will be cleaner fix. This bug was only reported on Apple mac during the full run  which was reason I went with prev fix. But open to change it to this if @aleksandr-voitenko concurs?
```
rescale.visible = (encoderCurrentValue != ENCODER_NVENC_H264_TEX);
rescaleRes.visible = (encoderCurrentValue != ENCODER_NVENC_H264_TEX);
```
[edit] This issue was discussed in slack. Went with this solution and updated PR

PR #1634, opened 2026-02-25
Consolidate Video Encoder Processing

mhoyer-streamlabs:
### Description
Consolidate video encoder processing into one place. Previously, encoder IDs, etc were hard coded multiple places leaving significant room for error. I created `osn-encoders.hpp/cpp` and included this anywhere these values are needed. `encoders_set`, containing encoder IDs and conversions, was renamed `videoEncoderOptions`  and moved into this file. It was also updated to include additional fields - `backup` to cover a case where a backup encoder is needed, `preset` to keep Preset names in a single place, and `family`  which is used in updating settings. Again, these values were previously hard coded multiple places. Additionally, quite a few encoders previously defaulted to obs_x264 in incomplete if/switch statements, those cases are now handled correctly. Other updates:

- Some encoders were not previously included in `encoders_set ` so those were added with the appropriate values
- Added helper functions to retrieve values from `videoEncoderOptions` rather than hard coding values in `if` or `switch` statements
- Separate logic of finding available encoders from creating a list so that we can use the same functionality to validate a single encoder vs returning an entire list
- New API - validate recording encoders on creation (since it is based on a flag that doesn't change), validate streaming encoders on start (because it is based on a service and we are certain to have all necessary info on start but not creation)
- Old and new API - if simple mode, convert the simple encoder name to the internal encoder name on creation so the FE doesn't have to know about encoder conversions
- Renamed a couple things for clarity

I also made some FE changes to utilize this new flow (PR coming soon). Previously, the FE would convert encoders in a couple different ways before sending them back to the BE, and this caused errors in several cases because hard coded values didn't match. Remove these conversions and allow the BE to handle all of it. 

### Motivation and Context
Minimize the chance of bugs stemming from encoder information being hard coded in multiple places.

### How Has This Been Tested?
Tested locally with all available encoders, both recording and streaming. 

### Types of changes

### Checklist:


obs-studio-server/source/osn-simple-recording.cpp
```diff
@@ -621,22 +627,16 @@ void osn::ISimpleRecording::SetLegacySettings(void *data, const int64_t id, cons
...
-	if (recording->videoEncoder) {
-		const char *encId = nullptr;
+	//don't save the encoder if using the streaming encoder, set it to empty string in that case
+	if (recording->quality != RecQuality::Stream && recording->videoEncoder) {
 		const char *encIdOBS = obs_encoder_get_id(recording->videoEncoder);
-		if (strcmp(encIdOBS, ADVANCED_ENCODER_X264) == 0 && !recording->lowCPU) {
-			encId = SIMPLE_ENCODER_X264;
-		} else if (strcmp(encIdOBS, ADVANCED_ENCODER_X264) == 0 && recording->lowCPU) {
+		std::string encId = osn::EncoderUtils::getSimpleEncoderFromInternal(encIdOBS);
+		if ((encId == SIMPLE_ENCODER_X264) == 0 && recording->lowCPU) {
```
aleksandr-voitenko: `lowCPU` save logic is inverted. This condition sets x264_lowcpu when encoder is **not** x264.
Likely, this is some slip after usage of strcmp. This should be written as
```
if (recording->lowCPU && encId == SIMPLE_ENCODER_X264) {
```

obs-studio-server/source/osn-encoders.hpp
```diff
@@ -0,0 +1,209 @@
...
+	bool check_availability;
+	bool check_availability_streaming;
+	bool check_availability_format;
+	bool only_for_reuse_simple;
+	std::string preset;
+	std::string family;
+	const std::string getSimpleName() const { return simple_internal_name.empty() ? simple_name : simple_internal_name; }
+};
+
+static std::vector<EncoderSettings> videoEncoderOptions = {
```
aleksandr-voitenko: Please, move the definition to the `cpp` file.
You can make it available everywhere by removing static and declaring as
```c++
extern const std::vector<EncoderSettings> videoEncoderOptions;
```

obs-studio-server/source/osn-multitrack-video-output.cpp
```diff
@@ -65,8 +65,8 @@ static void adjust_video_encoder_scaling(const obs_video_info &ovi, obs_encoder_
 	obs_encoder_set_scaled_size(video_encoder, requested_width, requested_height);
 	obs_encoder_set_gpu_scale_type(video_encoder, encoder_config.gpu_scale_type.value_or(OBS_SCALE_BICUBIC));
 	obs_encoder_set_preferred_video_format(video_encoder, encoder_config.format.value_or(VIDEO_FORMAT_NV12));
-	obs_encoder_set_preferred_color_space(video_encoder, encoder_config.colorspace.value_or(VIDEO_CS_709));
-	obs_encoder_set_preferred_range(video_encoder, encoder_config.range.value_or(VIDEO_RANGE_PARTIAL));
+	//obs_encoder_set_preferred_color_space(video_encoder, encoder_config.colorspace.value_or(VIDEO_CS_709));
+	//obs_encoder_set_preferred_range(video_encoder, encoder_config.range.value_or(VIDEO_RANGE_PARTIAL));
```
aleksandr-voitenko: Please, return back this code. In order to make it work, you need to pull the latest OBS version from the `streamlabs/obs` repository.

obs-studio-server/source/osn-encoders.cpp
```diff
@@ -0,0 +1,443 @@
...
+}
+
+std::string osn::EncoderUtils::getSimpleEncoderFromInternal(const char *encoder)
+{
+	//this defaults to advanced b/c that's how it's done in osn-simple-streaming where this is used....WHY
+	std::string encoderName = ADVANCED_ENCODER_X264;
+	bool found = false;
+
+	for (const auto curEnc : videoEncoderOptions) {
+		if (encoder == curEnc.simple_internal_name) {
```
aleksandr-voitenko: Please correct me if I'm wrong here.
Current forward mapping in `getInternalEncoderFromSimple` supports a fallback via backup:
- For simple nvenc, it prefers obs_nvenc_h264_tex
- If that isn’t registered, it falls back to ffmpeg_nvenc (backup)

But reverse mapping in `getSimpleEncoderFromInternal` only checks simple_internal_name, not backup.
So when encoder id is ffmpeg_nvenc, reverse mapping misses and falls back to x264.

Why this is bad:
On systems where obs_nvenc_h264_tex is unavailable but ffmpeg_nvenc is available, opening/saving settings can silently rewrite encoder selection to x264.

Compared to previous behavior (before this refactor):
- Mapping was done with explicit branches, and ADVANCED_ENCODER_NVENC (ffmpeg_nvenc) was explicitly mapped back to simple NVENC.
- So the fallback path round-tripped correctly and did not collapse to x264.

Proposed fix (in this line):
```c++
if ((encoder == curEnc.simple_internal_name) || (!curEnc.backup.empty() && encoder == curEnc.backup)) {
```

obs-studio-server/source/nodeobs_settings.cpp
```diff
@@ -1301,17 +1075,16 @@ void OBS_settings::getSimpleOutputSettings(std::vector<SubCategory> *outputSetti
...
 			preset.push_back({"Speed", ipc::value("speed")});
 			preset.push_back({"Balanced", ipc::value("balanced")});
 			preset.push_back({"Quality", ipc::value("quality")});
 			entries.push_back(preset);
 			defaultPreset = "balanced";
-
-		} else if (encoder == SIMPLE_ENCODER_NVENC || encoder == ADVANCED_ENCODER_NVENC || encoder == ENCODER_NVENC_H264_TEX ||
-			   encoder == ENCODER_NVENC_HEVC_TEX) {
-			preset = createSettingEntry("NVENCPreset2", "OBS_PROPERTY_LIST", "Encoder Preset (higher = less CPU)", "OBS_COMBO_FORMAT_STRING");
+		} else if (presetName == PRESET_NVENC) {
```
aleksandr-voitenko, line 1074: This code almost certainly introduces a regression.
#### Why this is a bug now.
In the new metadata-driven flow:
- `getEncoderPreset` returns EncoderSettings.preset
- The code in this line branches on that value, if presetName == PRESET_NVENC ("NVENCPreset2").
- But NVENC rows in `videoEncoderOptions` still store literal "NVENCPreset" (legacy key).

So for NVENC:
- `getEncoderPreset` returns "NVENCPreset" (legacy),
- presetName == PRESET_NVENC fails,
- NVENC preset branch is skipped in simple output settings

#### Previous (pre-refactor):

`getSimpleOutputSettings()` used explicit encoder checks (if encoder == NVENC...) and directly created "NVENCPreset2" UI entry.
So NVENC always took the NVENC preset path there.


#### Proposal
- Make NVENC metadata use PRESET_NVENC consistently. You can update `videoEncoderOptions`.
- Keep backward compatibility for old configs in `nodeobs_service.cpp` by reading legacy NVENCPreset when NVENCPreset2 is missing.

obs-studio-server/source/nodeobs_settings.cpp
```diff
@@ -3008,8 +2782,16 @@ void OBS_settings::saveAdvancedOutputRecordingSettings(std::vector<SubCategory>
 	int ret = config_save_safe(ConfigManager::getInstance().getBasic(), "tmp", nullptr);
 
 	if (newEncoderType) {
+		//this is called immediately on encoder change so no other settings have been changed - start with defaults
 		encoderSettings = obs_encoder_defaults(config_get_string(ConfigManager::getInstance().getBasic(), section.c_str(), "RecEncoder"));
-		OBS_service::createVideoRecordingEncoder();
+
+		//this defaults to obs_x264 so create the correct encoder with default settings - getSettings called immediately after and will update it correctly but
+		//do it right here just in case
+		//OBS_service::createDefaultSimpleVideoRecordingEncoder();
```
aleksandr-voitenko: Seems like this should be removed.

obs-studio-server/source/osn-recording.cpp
```diff
@@ -143,8 +150,12 @@ obs_encoder_t *osn::IRecording::duplicate_encoder(obs_encoder_t *src, uint64_t t
 
 	if (obs_encoder_get_type(src) == OBS_ENCODER_AUDIO) {
 		dst = obs_audio_encoder_create(obs_encoder_get_id(src), name.c_str(), obs_encoder_get_settings(src), trackIndex, nullptr);
+		//TODO added so it eventually gets released...is this correct? dont' even include osn-audio-encoder here
+		osn::AudioEncoder::Manager::GetInstance().allocate(dst);
```
aleksandr-voitenko, line 158: > so it eventually gets released

During shutdown, at the moment. Manager does not own the object. We have very similar anti-patterns in other places, so it must be fixed altogether and not in this PR.

obs-studio-server/source/osn-simple-recording.cpp
```diff
@@ -317,16 +317,20 @@ void osn::SimpleRecording::UpdateEncoders()
 		videoEncoder = streaming->videoEncoder;
 		audioEncoder = streaming->audioEncoder;
 		if (obs_get_multiple_rendering()) {
-			obs_encoder_t *videoEncDup = osn::IRecording::duplicate_encoder(videoEncoder);
-			videoEncoder = videoEncDup;
+			//TODO this doesn't register with the manager, should it?
```
aleksandr-voitenko: Only if we need to have a handle in the IPC API. So likely yes, we need it.

obs-studio-server/source/osn-simple-streaming.cpp
```diff
@@ -277,6 +278,7 @@ void osn::SimpleStreaming::UpdateEncoders()
 	int aBitrate = static_cast<int>(obs_data_get_int(audioEncSettings, "bitrate"));
 
 	std::string id = obs_encoder_get_id(videoEncoder);
+	//TODO why just AMD here?
```
aleksandr-voitenko: I believe it was copied from OBS UI code.

aleksandr-voitenko (commented):
Overall. Too many `TODO:`s and unanswered questions for a production-ready solution.

obs-studio-server/source/osn-simple-recording.cpp
mhoyer-streamlabs: Good catch, definitely left from strcmp - fixed.

obs-studio-server/source/osn-multitrack-video-output.cpp
mhoyer-streamlabs: Included accidentally - fixed.

obs-studio-server/source/osn-encoders.cpp
mhoyer-streamlabs: Fixed.

obs-studio-server/source/nodeobs_settings.cpp
mhoyer-streamlabs: I can't believe I defined the presets and didn't actually use them in `videoEncoderOptions`. I'll go sit in the corner now.

obs-studio-server/source/osn-recording.cpp
mhoyer-streamlabs: I'm going to leave in these allocate statements on `duplicate_encoder` so the encoders can be released on shutdown. The lack of a `free` call is present for all encoders so, as you said, that piece will be handled later.

aleksandr-voitenko (approved):

obs-studio-server/source/osn-simple-streaming.cpp
```diff
@@ -533,7 +544,7 @@ void osn::ISimpleStreaming::GetLegacySettings(void *data, const int64_t id, cons
 
 void osn::ISimpleStreaming::SetLegacyVideoEncoderSettings(obs_encoder_t *encoder)
 {
-	const char *encId = nullptr;
+	//const char *encId = nullptr;
```
sandboxcoder: looks like we can remove this

sandboxcoder (approved):

PR #1635, opened 2026-03-06
Fix encoder vs output shutdown crash

aleksandr-voitenko:
### Description
This PR reduces shutdown-time crashes in `obs-studio-node` by making shutdown wait for all active outputs to fully stop before cleanup continues. The change is made in `nodeobs_service.cpp` and is intended to avoid races where streaming/recording resources are destroyed while OBS output threads are still active.

The stack trace from Sentry. See, for example an issue with ID `2a0f8364`
```
w32-pthreads +0x05277   pthread_mutex_unlock (pthread_mutex_unlock.c:61)
KERNELBASE   +0xc37ea   ResetEvent
obs          +0x6439d   gpu_encode_thread (obs-video-gpu-encode.c:126)
```

The stack trace is not really meaningful, but logs associated with the issue are the real source of insight.

### What changed
-   Improved  stopStreaming(...)  to handle  nullptr  outputs safely and to treat  connecting  as a busy state instead of assuming only  active/reconnecting  matter.
-   Added shutdown helpers that detect whether an output is still busy and block shutdown until all relevant outputs have stopped.
-   Updated  stopAllOutputs()  to:
    -   wait for any prior release worker to finish,
    -   stop stream outputs gracefully,
    -   stop replay buffer, recording, and virtual camera outputs if still busy,
    -   wait until all outputs are no longer active/connecting/reconnecting,

### How Has This Been Tested?
Manually, Windows only.

### Types of changes

### DIFF
--- a/obs-studio-server/source/nodeobs_service.cpp
+++ b/obs-studio-server/source/nodeobs_service.cpp
@@ -3347,7 +3347,9 @@ void WaitForAllOutputsToStop()
 
 			blog(LOG_ERROR, "%s", crashMessage.c_str());
 			util::CrashManager::AddWarning(crashMessage);
+#ifdef WIN32
 			util::CrashManager::GetMetricsProvider()->BlameServer();
+#endif
 
 			std::terminate();
 		}
--- a/obs-studio-server/source/nodeobs_service.cpp
+++ b/obs-studio-server/source/nodeobs_service.cpp
@@ -3283,11 +3283,8 @@ bool OutputIsBusy(obs_output_t *output, bool includeReconnect = false)
 void WaitForAllOutputsToStop()
 {
 	// At the moment, we assume that all outputs stop eventually, so there is no deadline timeout.
-	while (OutputIsBusy(streamingOutput[StreamServiceId::Main], true) ||
-	       OutputIsBusy(streamingOutput[StreamServiceId::Second], true) ||
-		   OutputIsBusy(recordingOutput) ||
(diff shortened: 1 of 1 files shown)


sandboxcoder (approved):

obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -3274,19 +3282,70 @@ void OBS_service::OBS_service_updateVirtualCam(void *data, const int64_t id, con
...
+
+	if (obs_output_active(output) || obs_output_connecting(output))
+		return true;
+
+	return includeReconnect && obs_output_reconnecting(output);
+}
+
+void WaitForAllOutputsToStop()
+{
+	// At the moment, we assume that all outputs stop eventually, so there is no deadline timeout.
```
summeroff: When a process is already in a shutdown sequence, it is safer to crash than to leave a zombie process infinitely waiting for output to close. We cannot be sure that all outputs will stop every time within a reasonable time.

obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -3259,19 +3267,70 @@ void OBS_service::OBS_service_updateVirtualCam(void *data, const int64_t id, con
...
```
summeroff: When a process is already in a shutdown sequence, it is safer to crash than to leave a zombie process infinitely waiting for output to close. We cannot be sure that all outputs will stop every time within a reasonable time.

summeroff (approved):

PR #1656, opened 2026-04-02
multitrack: restore encoder gpu scale

sandboxcoder:
### Description
* revert preferred colorspace to `VIDEO_CS_709` which was a change I made in ac570163 in which did not address root issue after more investigation
* removing the redundant (?) `obs_encoder_set_video_mix` call. This was already set by the `create_video_encoder()` right above this line so not sure why we need to call again?

### Motivation and Context
Enable enhanced broadcasting to at least display. Still investigating delayed audio

### How Has This Been Tested?
Verified all encoders (1080, 720p, etc) all display properly when I stream to Twitch during enhanced broadcasting.

### Types of changes

### Checklist:


sandboxcoder:
closed (new PR posted here)- https://github.com/streamlabs/obs-studio/pull/720 that will redo how the some of the BE logic works

PR #1664, opened 2026-04-08
multitrack: revert encoder gpu scale

sandboxcoder:
### Description
* revert preferred colorspace to VIDEO_CS_709 which I recently changed in
commit ac57016. But after diagnosing the root cause properly, that change was
[unnecessary](https://github.com/streamlabs/obs-studio/pull/720).
* removing the redundant (?) `obs_encoder_set_video_mix` call. This was already set by the `create_video_encoder()` right above this line so not sure why we need to call again? It does not cause any issues but enhanced broadcasting appears to work without this call?

### Motivation and Context
Optional cleanup

### How Has This Been Tested?
Verified all encoders (1080, 720p, etc) all display properly when I stream to Twitch during enhanced broadcasting.

### Types of changes

### Checklist:

### DIFF
--- a/obs-studio-server/source/osn-multitrack-video-output.cpp
+++ b/obs-studio-server/source/osn-multitrack-video-output.cpp
@@ -65,7 +65,7 @@ static void adjust_video_encoder_scaling(const obs_video_info &ovi, obs_encoder_
 	obs_encoder_set_scaled_size(video_encoder, requested_width, requested_height);
 	obs_encoder_set_gpu_scale_type(video_encoder, encoder_config.gpu_scale_type.value_or(OBS_SCALE_BICUBIC));
 	obs_encoder_set_preferred_video_format(video_encoder, encoder_config.format.value_or(VIDEO_FORMAT_NV12));
-	obs_encoder_set_preferred_color_space(video_encoder, encoder_config.colorspace.value_or(VIDEO_CS_DEFAULT));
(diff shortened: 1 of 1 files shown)


obs-studio-server/source/osn-multitrack-video-output.cpp
```diff
@@ -209,8 +209,6 @@ static bool create_video_encoders(const Config &go_live_config, std::shared_ptr<
 
 		if (obs_get_multiple_rendering()) {
 			obs_encoder_set_video_mix(encoder, obs_video_mix_get(ovi, OBS_STREAMING_VIDEO_RENDERING));
-		} else {
```
mhoyer-streamlabs, line 212: This needs to remain - multiple rendering is used for selective recording; the last param is what differentiates.

mhoyer-streamlabs: Actually I misunderstood why you removed it - you're saying it's done elsewhere before this.

sandboxcoder: Thanks, updated

sandboxcoder: All good, I will put it back I just saw your latest post

mhoyer-streamlabs (approved):

PR #1667, opened 2026-04-09
Update client error handling

mhoyer-streamlabs:
### Description

Update client error handling to ensure all errors are passed along to the front end by replacing `conn->call` with `conn->call_synchronous_helper` and parsing the return. In some streaming/recording/replay buffer cases, clear the existing object ref and don't set the new one if the back end call fails. Replaced `std::vector<ipc::value>` with `auto` to shorten calls and make them easier to read.

### Motivation and Context

Some important errors were being ignored.

### Types of changes

### Checklist:

### DIFF
--- a/obs-studio-client/source/advanced-replay-buffer.cpp
+++ b/obs-studio-client/source/advanced-replay-buffer.cpp
@@ -196,14 +196,13 @@ Napi::Value osn::AdvancedReplayBuffer::GetStreaming(const Napi::CallbackInfo &in
 
 void osn::AdvancedReplayBuffer::SetStreaming(const Napi::CallbackInfo &info, const Napi::Value &value)
 {
-	if (!parentOutputRef.IsEmpty())
-		parentOutputRef.Reset();
-
 	auto conn = GetConnection(info);
 	if (!conn)
 		return;
 
 	if (value.IsNull() || value.IsUndefined()) {
+		if (!parentOutputRef.IsEmpty())
+			parentOutputRef.Reset();
 		auto response = conn->call_synchronous_helper(className, "SetStreaming", {ipc::value(this->uid), ipc::value(UINT64_MAX)});
 		ValidateResponse(info, response);
 		usesStream = false;
@@ -226,6 +225,8 @@ void osn::AdvancedReplayBuffer::SetStreaming(const Napi::CallbackInfo &info, con
 		return;
 
 	usesStream = true;
+	if (!parentOutputRef.IsEmpty())
+		parentOutputRef.Reset();
 	parentOutputRef = Napi::Persistent(obj);
 }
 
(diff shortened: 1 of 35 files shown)


obs-studio-client/source/advanced-replay-buffer.cpp
```diff
@@ -237,15 +240,17 @@ Napi::Value osn::AdvancedReplayBuffer::GetRecording(const Napi::CallbackInfo &in
...
 		return;
 
 	if (value.IsNull() || value.IsUndefined()) {
-		if (!parentOutputRef.IsEmpty())
-			parentOutputRef.Reset();
-		conn->call(className, "SetRecording", {ipc::value(this->uid), ipc::value(UINT64_MAX)});
-		usesStream = false;
+		auto response = conn->call_synchronous_helper(className, "SetRecording", {ipc::value(this->uid), ipc::value(UINT64_MAX)});
+		ValidateResponse(info, response);
+		usesStream = true;
```
aleksandr-voitenko: Why did you change `usesStream = true;` ? This changes the behavior entirely. Likely a copypaste issue.

obs-studio-client/source/simple-replay-buffer.cpp
```diff
@@ -213,15 +215,17 @@ Napi::Value osn::SimpleReplayBuffer::GetRecording(const Napi::CallbackInfo &info
...
 		return;
 
 	if (value.IsNull() || value.IsUndefined()) {
-		if (!parentOutputRef.IsEmpty())
-			parentOutputRef.Reset();
-		conn->call(className, "SetRecording", {ipc::value(this->uid), ipc::value(UINT64_MAX)});
-		usesStream = false;
+		auto response = conn->call_synchronous_helper(className, "SetRecording", {ipc::value(this->uid), ipc::value(UINT64_MAX)});
+		ValidateResponse(info, response);
+		usesStream = true;
```
aleksandr-voitenko: `usesStream = true;`
The same here.

obs-studio-client/source/streaming.cpp
```diff
@@ -121,14 +122,16 @@ Napi::Value osn::Streaming::GetVideoEncoder(const Napi::CallbackInfo &info)
 
 void osn::Streaming::SetVideoEncoder(const Napi::CallbackInfo &info, const Napi::Value &value)
 {
+	if (!videoEncoderRef.IsEmpty())
```
aleksandr-voitenko: Resetting the cached ref before the IPC call succeeds can leave the wrapper out of sync with the backend. If `SetVideoEncoder` rejects the new encoder, the backend keeps the old encoder but `videoEncoderRef` has already been cleared, so the client starts returning `undefined`. Can we move the reset until after `ValidateResponse` succeeds?

This same pattern also appears in other places.

mhoyer-streamlabs: You are correct - I was thinking a backend failure meant no encoder stored, but I see now that it actually keeps any previous value in both client and server. My thought was not to store the previous value in an error case, but I only updated half of that. I'll move those statements back down so the previous value remains.

obs-studio-client/source/advanced-replay-buffer.cpp
mhoyer-streamlabs: Definitely looks like copy/paste. Although should we even set this here? If setting the new recording object fails and we're going to leave the previous object, shouldn't we leave the previous value of this var as well?

sandboxcoder (approved):

summeroff (approved):

PR #1672, opened 2026-04-17
fix way to set sources for virtual cam

summeroff:
**What changed**

Reworked how Virtual Cam resolves and binds its output source (UpdateVirtualCamOutputSource + new helpers ResolveVCamScene/ResolveVCamSource).
- **SceneOutput**: now uses direct backstage activation on the resolved scene instead of copying items into a throwaway wrapper. No more stale content or item accumulation.
- **SourceOutput**: creates a private scene on-demand and centers/scales the source properly.
- **ProgramView**: falls back to the primary canvas (fixes multi-canvas builds where obs_get_video() is unavailable).
- **Added robust fallbacks**: if the configured scene/source name can't be found, it silently switches to the program scene (or first non-scene item in it).

Full cleanup overhaul (DestroyVirtualCamView, DestroyVirtualCameraScene, DeactivateSources):
- Fixed shutdown crashes and ref-count leaks.
- Removed double backstage deactivation.
- Explicitly release scene items.
- Proper webcam view for stopAllOutputs.
 

**Why**
The old path was fragile, produced duplicate items, leaked views, and crashed on shutdown or when switching output types. This makes Virtual Cam reliable again across all output modes.

### How Has This Been Tested?
Manual testing in app. 

### Types of changes

### Checklist:

### DIFF
--- a/obs-studio-server/source/nodeobs_service.cpp
+++ b/obs-studio-server/source/nodeobs_service.cpp
@@ -3200,12 +3200,15 @@ void OBS_service::DestroyVirtualCameraScene()
 
 static obs_video_info *GetPrimaryVCamCanvas()
 {
-	obs_video_info *primary = nullptr;
-	osn::Video::Manager::GetInstance().for_each([&primary](obs_video_info *canvas) {
-		if (!primary)
-			primary = canvas;
+	obs_video_info *best = nullptr;
+	osn::Video::Manager::GetInstance().for_each([&best](obs_video_info *canvas) {
+		if (!best || canvas->base_width * best->base_height > best->base_width * canvas->base_height)
+			best = canvas;
 	});
-	return primary;
+	if (best)
+		return best;
+
+	return videoInfo[StreamServiceId::Main];
 }
 
 // Resolve the scene to render for SceneOutput. Falls back to the current program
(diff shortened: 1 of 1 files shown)


obs-studio-server/source/nodeobs_service.cpp
```diff
@@ -3170,19 +3174,88 @@ void OBS_service::DestroyVirtualCamView()
...
+	obs_deactivate_scene_on_backstage(sceneSource);
 	obs_scene_release(vCamSourceScene);
 	vCamSourceScene = nullptr;
-	vCamSourceSceneItem = nullptr;
+}
+
+static obs_video_info *GetPrimaryVCamCanvas()
+{
+	obs_video_info *primary = nullptr;
+	osn::Video::Manager::GetInstance().for_each([&primary](obs_video_info *canvas) {
```
aleksandr-voitenko: That is allocation-order based, not service-mapping based. I understand that this works, but the code is a little bit fragile.

Before this chunk I would suggest adding this
```c++
	if (videoInfo[StreamServiceId::Main])
		return videoInfo[StreamServiceId::Main];
```

aleksandr-voitenko (approved):

PR #1691, opened 2026-05-06
MacOS getDevice() optimization, set CI env var, & add getVideoDevices() test

sandboxcoder:
### Description
* Restore commit 9279b8a3 now that this has been fully root caused. Bringing back this change because it fixes the case where coreaudio spins up a `reconnect_thread` whenever Desktop queries for audio devices. This was because the previous implementation passed in a dummy device which causes coreaudio to create a thread to poll if this dummy device is later initialized. This will cleanup the logs eliminating this warning (32 occurrences in a log I checked from Desktop): `[Warning] [coreaudio_get_device_name] failed to get name: 560947818`
* Set CI env flag for mac test CI. This is so we can use the new `obs.isCI()` function to determine if we're running in a CI env (no GPU) very quickly from the tests.
* Adds a test for `OBS_settings_getVideoDevices`. 
* test_osn_advanced_replayBuffer.ts: still flaky on macOS test runner so
disable on CI runner
* Set `CMAKE_OSX_DEPLOYMENT_TARGET` to MacOS 12. now matches SLOBS which is set [here](https://github.com/streamlabs/obs-studio/blob/5def36cca41a99e24298e6ecd0b678d54dcec240/CMakePresets.json#L36)
 
Performance:
New implementation (nanoseconds)
```
[000:00:00:03.691.221.250][66014221][Info] OBS_settings_getInputAudioDevices took 400125
[000:00:00:03.691.641.333][66014221][Info] OBS_settings_getOutputAudioDevices took 288458
```
Old (nanoseconds)
```
[000:00:00:03.530.971.333][65989647][Info] OBS_settings_getInputAudioDevices took 721292
[000:00:00:03.531.601.833][65989647][Info] OBS_settings_getOutputAudioDevices took 479084
```
The new `enumDevices` implementation is more efficient since it does not create a `dummy_source`. Also, it avoids triggering coreaudio to create a reconnect_thread for a dummy device (which saves us another 2ms or so in an async thread because mac-coreaudio polls within the reconnect_thread). 

### Motivation and Context
When I was looking into profiling missing source devices it reminded me that getDevices() triggers coreaudio to create a reconnect_thread which does later get reaped however it did not seem ideal. But it would be nice to avoid this. Currently, desktop queries getAudioDevices() at least around 14 times so it is nice to optimize this to get the app to load up faster and spin up less threads. 

### How Has This Been Tested?
Locally, I verified the same audio/video devices are output. Originally, commit 9279b8a3 was reverted due to breaking the old onboarding flow (no webcam detected) but that is addressed in this PR since we will enumerate video devices using the legacy approach.

### Types of changes

### Checklist:


obs-studio-server/source/nodeobs_settings-osx.mm
```diff
@@ -0,0 +1,35 @@
...
+
+		AVCaptureDeviceDiscoverySession *session = [AVCaptureDeviceDiscoverySession discoverySessionWithDeviceTypes:deviceTypes
+														  mediaType:AVMediaTypeVideo
+														   position:AVCaptureDevicePositionUnspecified];
+
+		for (AVCaptureDevice *device in session.devices) {
+			if (!device.localizedName || !device.uniqueID)
+				continue;
+			result.push_back({[device.localizedName UTF8String], [device.uniqueID UTF8String]});
+		}
```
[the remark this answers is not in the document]
sandboxcoder: Interesting I may need to update the project setting at some point because our min target is macOS 12.

tests/osn-tests/util/obs_handler.ts
```diff
@@ -557,6 +557,12 @@ export class OBSHandler {
         return this.os === 'darwin';
     }
 
+    isOnDarwinCI()
+    {
+        // Wrapped this in a function- just incase we want to add more conditions later or disable only within the build agent.
+        return this.os === 'darwin' && this.ci;
```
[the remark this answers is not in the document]
sandboxcoder: @copilot apply changes based on [this feedback](https://github.com/streamlabs/obs-studio-node/pull/1691#discussion_r3198141855)

tests/osn-tests/util/obs_handler.ts
```diff
@@ -553,8 +553,8 @@ export class OBSHandler {
 
     isDarwin()
     {
-        // Wrapped this in a function- just incase we want to add more conditions later or disable only within the build agent.
-        return this.os === 'darwin';
+        // Wrapped this in a function- just in case we want to add more conditions later or disable only within the build agent.
+        return this.os === 'darwin' && this.ci;
     }
```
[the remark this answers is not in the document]
sandboxcoder: Will do this in a future PR. will touch many test files

obs-studio-server/source/nodeobs_settings.cpp
```diff
@@ -4097,9 +4075,17 @@ void OBS_settings::OBS_settings_getVideoDevices(void *data, const int64_t id, co
 	rval.push_back(ipc::value((uint32_t)0));
 	enumVideoDevices(rval);
 #elif __APPLE__
-	const char *source_id = "macos_avcapture";
-	const char *property_name = "device";
-	getDevices(source_id, property_name, rval);
+	// Here we do not invoke getDevices() because the mac-capture
+	// plugin will not enumerate video devices unless it is fully initialized.
+	// So instead, we will enumerate video devices manually. Hopefully,
+	// mac-capture doesn't do anything special to enumerate video devices.
```
aleksandr-voitenko: The upstream shows that the plugin does something special:
- It includes Desk View on macOS 13+
- It enumerates both `AVMediaTypeVideo` and `AVMediaTypeMuxed`
- The legacy path does the same kind of enumeration in `legacy/av-capture.mm`


Please see the comment above.

obs-studio-server/source/nodeobs_settings-osx.mm
```diff
@@ -0,0 +1,43 @@
+#include "nodeobs_settings-osx.h"
+
+#import <AVFoundation/AVFoundation.h>
+
+std::vector<std::pair<std::string, std::string>> getVideoDevicesMacOS()
```
aleksandr-voitenko: `OBS_settings_getVideoDevices()` may no longer match the real mac-avcapture source list.

Suggestion:
Please update `getVideoDevicesMacOS()` to mirror the paired mac-avcapture plugin enumeration exactly, including Desk View and muxed devices, or explain why `OBS_settings_getVideoDevices()` is intentionally allowed to return a narrower list than the source itself supports.

In my opinion (please correct me if I'm wrong), OSN must duplicate plugin enumeration from OBS, and add a short comment near the helper saying it must stay in sync with `plugins/mac-avcapture/plugin-properties.m`.

Implementation reference: `properties_update_device` function.

CMakeLists.txt
```diff
@@ -29,15 +29,15 @@ ENDIF()
 
 # Comfigure macos architecture and min version
 iF(APPLE)
-	set(CMAKE_XCODE_ATTRIBUTE_MACOSX_DEPLOYMENT_TARGET[arch=arm64] "11.0")
-	set(CMAKE_XCODE_ATTRIBUTE_MACOSX_DEPLOYMENT_TARGET[arch=x86_64] "10.15")
+	set(CMAKE_XCODE_ATTRIBUTE_MACOSX_DEPLOYMENT_TARGET[arch=arm64] "12.0")
+	set(CMAKE_XCODE_ATTRIBUTE_MACOSX_DEPLOYMENT_TARGET[arch=x86_64] "12.0")
 	if (NOT CMAKE_OSX_ARCHITECTURES)
```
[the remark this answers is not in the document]
sandboxcoder: Yes, Streamlabs Desktop minimum version is actually MacOS 12 (this version in particular is special, my minspec Intel laptop is set to MacOS 12 while the Mac M4 is using latest). I just never updated the deployment target in the xcode cmake file because we never needed to bump it (only reason I was thinking to bump it now is to make it obvious there is no need for fallback code for anything older then macOS 12) but should have bumped this much earlier...

This matches SLOBS Deployment [target now](https://github.com/streamlabs/obs-studio/blob/5def36cca41a99e24298e6ecd0b678d54dcec240/CMakePresets.json#L36).

obs-studio-server/source/nodeobs_settings.cpp
sandboxcoder: Ended up removing the custom video device enumeration so this is resolved. Just seems more efficient to use the slightly update `getDevicesUsingDummySource` which is the legacy approach. Made small modifications to it

obs-studio-server/source/nodeobs_settings-osx.mm
sandboxcoder: Ended up removing the custom video device enumeration so this is resolved.

obs-studio-server/source/nodeobs_settings.cpp
sandboxcoder: I realized I can keep using legacy implementation for this and it should also work for `macos_avcapture_fast` as well (because it shares same devices as macos_avcapture). 

Conclusion: we can keep iterating video devices using `macos_avcapture` plugin. I only need to update the coreaudio code path in this PR.

aleksandr-voitenko (approved):

PR #1695, opened 2026-05-12
gather data from user submitted long_call.txt files

sandboxcoder:
### Description
* Adds profile-long-calls subfolder to obs-studio-node.
* Analyzes the long_calls.txt file produced by obs-studio-node and helps identify which methods takes the longest.
* Asana: https://app.asana.com/1/1083097041131/project/1207748235152476/task/1214564097102925

### Motivation and Context
Check user submitted logs to AWS S3 bucket and generate a report.

### How Has This Been Tested?
Ran it manually over a 24 hour period.

### Types of changes

### Checklist:

### DIFF
--- a/profile-long-calls/long-calls-analyzer.ts
+++ b/profile-long-calls/long-calls-analyzer.ts
@@ -509,7 +509,10 @@ function printStats(db: DatabaseType): void {
   });
 
   console.log('\n');
-  cleanOldSources(db);
+  const cleanOld : boolean = process.argv.includes('--cleanOldSources');
+  if (cleanOld) {
+    cleanOldSources(db);
+  }
   printStats(db);
   db.close();
 })();
--- a/profile-long-calls/long-calls-analyzer.ts
+++ b/profile-long-calls/long-calls-analyzer.ts
@@ -72,16 +72,16 @@ interface ParsedKey {
 }
 
 /**
- * Cache keys are expected to use the format "{isoDate}-{username}.zip".
- * The ISO timestamp portion is always 24 characters
- * ("2024-01-15T10:23:45.123Z"), so the username begins at index 25.
- * Anonymous uploads omit the username suffix.
+ * Cache keys are formatted as "{isoDate}-{username}.zip" (see cache-uploader.ts).
+ * We match with a regex on the full key rather than path.basename so that usernames
+ * containing "/" (e.g. "foo /bar") are captured correctly.
  */
+const KEY_RE = /(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z)-?(.*)\.zip$/;
+
 function parseKey(key: string): ParsedKey {
-  const name = path.basename(key, '.zip');
-  const uploadedAt = new Date(name.substring(0, 24));
-  const username = name.length > 25 ? name.substring(25) : null;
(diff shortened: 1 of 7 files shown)


profile-long-calls/long-calls-analyzer.ts
```diff
@@ -0,0 +1,474 @@
...
+}
+
+// ─── Download + parse ─────────────────────────────────────────────────────────
+
+async function downloadToBuffer(key: string): Promise<Buffer> {
+  const resp = await s3.send(new GetObjectCommand({ Bucket: BUCKET, Key: key }));
+  const chunks: Uint8Array[] = [];
+  for await (const chunk of resp.Body as AsyncIterable<Uint8Array>) chunks.push(chunk);
+  return Buffer.concat(chunks);
+}
```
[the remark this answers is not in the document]
sandboxcoder: I am not sure if this is needed atm? Did not implement

profile-long-calls/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "dependencies": {
+    "@aws-sdk/client-s3": "^3.967.0",
+    "adm-zip": "^0.5.17",
+    "better-sqlite3": "^11.10.0"
+  },
+  "devDependencies": {
+    "@types/adm-zip": "^0.5.7",
+    "@types/better-sqlite3": "^7.6.13",
+    "@types/node": "^22.0.0",
```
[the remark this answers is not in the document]
sandboxcoder: I keep node 22 here because AWS dependency actually prints a warning if you're not using Node 22.

I updated README.md to indicate Node 22 is ideal

profile-long-calls/long-calls-analyzer.ts
```diff
@@ -0,0 +1,483 @@
...
+
+type ProcessResult = 'ok' | 'skipped' | 'no_long_calls' | 'bad_zip' | 'bad_entry';
+
+async function processKey(key: string, db: DatabaseType, stmts: Stmts): Promise<ProcessResult> {
+  if (db.prepare('SELECT 1 FROM sources WHERE s3_key = ?').get(key)) return 'skipped';
```
aleksandr-voitenko, line 233: This only skips the exact same S3 key, so resumed runs can still count older uploads from the same user. 

`deduplicateByUser()` keeps the latest cache per user only within the current S3 listing, but previously inserted `sources`/`long_calls` rows remain in `long_calls.db`, and `printStats()` aggregates across all rows. If a user has one cache inserted Monday and a newer cache inserted Tuesday, both are included in the report. 

Please either delete older rows for the same username when inserting a newer cache, make the report query select only the latest source per user, or require/reset the DB for each report.

summeroff: Repeated uploads from the same user are expected. But rare edge cases include: the user clearing their cache, resulting in an upload containing only new long_calls, or the user logging in with a different account while the local long_calls. So simple dropping old might be not a correct way. Individual long calls may be easier to identify by timestamp and pid.

sandboxcoder: Added optional `--cleanOldSources` flag. Not sure what is ideal solution here? I might have misunderstood? Added commit 81b3a13c and 6c2d28d7

PR #1699, opened 2026-05-19
Bind video mix before updating encoder

summeroff:
SimpleStreaming::updateEncoders() called obs_encoder_update(videoEncoder, ...) before obs_encoder_set_video_mix(videoEncoder, ...). On a stale libobs, that runs the encoder's update callback against a freed
  video_t and crashes (obs_encoder_get_width → dangling encoder->media). Reorders the calls so the encoder is rebound to a live mix before its update runs — defense in depth on top of the libobs fix.

### Types of changes

### Checklist:

### DIFF
--- a/obs-studio-server/source/osn-advanced-streaming.cpp
+++ b/obs-studio-server/source/osn-advanced-streaming.cpp
@@ -401,14 +401,14 @@ void osn::AdvancedStreaming::UpdateEncoders()
 		obs_encoder_set_preferred_video_format(videoEncoder, VIDEO_FORMAT_NV12);
 	}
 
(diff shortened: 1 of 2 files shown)


aleksandr-voitenko:
Advanced streaming still has the ordering that the PR fixes in `SimpleStreaming::UpdateEncoders()`: the encoder update callback can run before the video encoder is rebound to the current live mix. If stale `encoder->media` is the crash reason, `AdvancedStreaming` can likely hit the same issue. Can we move the `obs_encoder_set_video_mix(...)` block before this `obs_encoder_update(...)` call as well?

aleksandr-voitenko (approved):

PR #1708, opened 2026-05-22
 Bound Sentry crash-report buffers and drop dead code

summeroff:
Description
  ## Summary
  - Cap all crash-report buffers — `OBS log general` (150), `Server warnings` (50), `Last actions` (50). Two were previously unbounded.
  - Drop unused `OBSLogType::Errors/Warnings` paths and the never-called breadcrumbs code; rename ambiguous `warnings` → `serverWarnings` so it doesn't collide with the libOBS log warning stream.

### Types of changes

### Checklist:

### DIFF
--- a/obs-studio-server/source/util-crashmanager.cpp
+++ b/obs-studio-server/source/util-crashmanager.cpp
@@ -718,6 +718,10 @@ void util::CrashManager::HandleCrash(const std::string &_crashInfo, bool callAbo
 	} catch (...) {
 	}
 
+	// Annotations attached to the Sentry minidump:
+	//   "OBS log general" — rolling tail of the libOBS log (capped at LogReport::MaximumMessages lines)
+	//   "Last actions"    — recent IPC calls received by the server (capped at MaximumActionsRegistered)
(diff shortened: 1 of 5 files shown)


obs-studio-server/source/util-crashmanager.cpp
```diff
@@ -1021,51 +1025,21 @@ void RewindCallStack()
...
-	case OBSLogType::General: {
-		auto &general = OBS_API::getOBSLogGeneral();
-		while (!general.empty()) {
-			result.push_back(general.front());
-			general.pop();
-		}
-
-		break;
-	}
+	auto &general = OBS_API::getOBSLogGeneral();
```
aleksandr-voitenko: `RequestOBSLog()` used `OBS_API::logReport.general` through a mutable reference without synchronizing with `node_obs_log()`, which appends to the same deque under `logMutex`. That leaves a data race/UB risk during crash annotation collection. Please move this behind an `OBS_API` helper that snapshots and optionally clears the log while holding `logMutex`, preferably with `try_lock` behavior for crash-time use.

PR #1714, opened 2026-05-28
Fix Twitch VOD Track audio routing in advanced + EB modes

summeroff:
## Summary

Regression introduced in 1.21's migration to the factory-API streaming path: the user-configurable Twitch VOD Track silently sent the wrong audio (or no separate VOD audio at all) to Twitch's VOD output, depending on streaming mode.

Tech support reported users on 1.21.1 seeing audio they'd muted on the VOD Track still present in their Twitch VOD. Two distinct root causes, fixed together since they share one user-visible symptom.

## Root cause

**1. Advanced streaming - hardcoded mixer + 0/1-based field confusion.**
`osn-advanced-streaming.cpp::SetupTwitchSoundtrackAudio` was a fossilized copy of the legacy "Soundtrack by Twitch" plugin code:
- The VOD encoder was created with `kSoundtrackArchiveTrackIdx = 5` regardless of the user's track choice, so VOD audio always came from mixer 5 (UI Track 6).
- `audioTrackConfigs[streaming->twitchTrack]` used raw array indexing instead of `GetTrackConfig(twitchTrack)`, mismatching the 1-based contract honored by every other consumer in `osn-audio-track.cpp`.
- It also silently mutated default desktop sources' mixer masks (carry-over from the Soundtrack plugin) - actively wrong once the VOD mixer is user-chosen.

The legacy `GetLegacySettings`/`SetLegacySettings` stored the field as 0-based with `-1`/`+1` conversions, while the new factory-API setters stored 1-based values from the desktop UI. Same field, two conventions, depending on entry point.

**2. Enhanced Broadcasting advanced - VOD track silently disabled.**
`osn-enhanced-broadcasting-advanced-streaming.cpp::Start` checked `streaming->twitchVODSupported && streaming->enableTwitchVOD` to decide whether to request a VOD track from the multitrack output, but **never set `twitchVODSupported`**. The field defaulted to `false` (from `osn-streaming.hpp:44`); regular `AdvancedStreaming::Start` computes it via `isTwitchVODSupported()` before `SetupTwitchSoundtrackAudio`, but the EB Start skipped that.

Net effect with EB on: `vod_track_mixer = nullopt` -> multitrack output's `create_audio_encoders` early-returns at the VOD branch -> no VOD encoder ever attached -> Twitch's go-live POST body went out with `"vod_track_audio": false` -> Twitch returned no VOD audio configuration -> VOD silently used the main stream audio.

## Fix

**`osn-advanced-streaming.cpp`**
- Route the VOD encoder through `GetMixerIndex(streaming->twitchTrack)` (user's 1-based choice -> 0-based mixer).
- Look the bitrate config up via `GetTrackConfig(streaming->twitchTrack)`; bail cleanly if the slot has no AudioTrack.
- Recreate the encoder when the mixer index would change between streams (mixer is fixed at `obs_audio_encoder_create` time).
- Drop the silent mutation of default desktop sources' mixer masks (and the corresponding restore in `StopTwitchSoundtrackAudio`).
- Normalize `GetLegacySettings`/`SetLegacySettings` to store `twitchTrack`/`audioTrack` as 1-based, matching the constructor defaults and the new API path - eliminates the two-conventions-in-one-field footgun.
- Dead-code removal: `kSoundtrackArchiveTrackIdx`, `setMixer()` helper.

**`osn-enhanced-broadcasting-advanced-streaming.cpp`**
- Compute `streaming->twitchVODSupported = streaming->isTwitchVODSupported()` when `enableTwitchVOD` is true, mirroring the regular path.
- Convert `streaming->twitchTrack` to a 0-based mixer index via `GetMixerIndex` before passing to `StartEnhancedBroadcastingStream` (which forwards to multitrack output as a raw mixer_idx).

## Test plan

- [x] **Regular advanced + VOD enabled, VOD Track = 3.** Log shows `'Twitch VOD Track Encoder'` created with `track: 3` (mixer 2). Audio muted on Track 3 via Advanced Audio Properties is excluded from VOD as expected.
- [x] **Enhanced Broadcasting + VOD enabled.** Log shows `'multitrack video vod audio 0'` encoder created; Go-live POST body contains `"vod_track_audio": true`. VOD receives the user-configured mix.
- [x] **Same scenarios with EB toggled off then on again** - both modes route VOD audio correctly without restart.
- [ ] **Twitch Dual Streaming (single connection, both canvases) + VOD.** Same EB code path as scenario 2; covered transitively.

## Out of scope (follow-ups)

- `osn-simple-streaming.cpp::SetupTwitchSoundtrackAudio` still has the hardcoded mixer 5 pattern, but the VOD Track UI is gated to advanced output mode in `GlobalSettings.tsx` so it's unreachable from a normal desktop session. Separate cleanup if simple-mode VOD is ever wired up.
- `streamArchive` / `SetupTwitchSoundtrackAudio` are misnomers now (no Soundtrack plugin involvement) and `oldMixer_desktopSource1/2` on `osn-streaming.hpp` are unused by the advanced path. Pure rename, separate PR.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/obs-studio-server/source/osn-enhanced-broadcasting-advanced-streaming.cpp
+++ b/obs-studio-server/source/osn-enhanced-broadcasting-advanced-streaming.cpp
@@ -118,7 +118,19 @@ void osn::IEnhancedBroadcastingAdvancedStreaming::Start(void *data, const int64_
 		PRETTY_ERROR_RETURN(ErrorCode::InvalidReference, "Invalid service.");
 	}
 
-	auto vod_track_mixer = (streaming->twitchVODSupported && streaming->enableTwitchVOD) ? std::optional{streaming->twitchTrack} : std::nullopt;
+	// twitchVODSupported is computed lazily at stream start (depends on the
+	// current service). The regular AdvancedStreaming::Start does this same
+	// check before SetupTwitchSoundtrackAudio; mirror it here so the VOD
+	// encoder is actually requested when the user enables VOD Track.
+	if (streaming->enableTwitchVOD) {
+		streaming->twitchVODSupported = streaming->isTwitchVODSupported();
+	}
+
+	// streaming->twitchTrack is a 1-based track number; the multitrack output
+	// expects a 0-based mixer index for vod_track_mixer.
+	auto vod_track_mixer = (streaming->twitchVODSupported && streaming->enableTwitchVOD)
+				       ? std::optional<size_t>{osn::IAudioTrack::GetMixerIndex(streaming->twitchTrack)}
+				       : std::nullopt;
 	try {
 		streaming->StartEnhancedBroadcastingStream(vod_track_mixer);
 	} catch (const std::exception &error) {
--- a/obs-studio-server/source/osn-enhanced-broadcasting-advanced-streaming.cpp
+++ b/obs-studio-server/source/osn-enhanced-broadcasting-advanced-streaming.cpp
@@ -126,11 +126,14 @@ void osn::IEnhancedBroadcastingAdvancedStreaming::Start(void *data, const int64_
 		streaming->twitchVODSupported = streaming->isTwitchVODSupported();
 	}
 
-	// streaming->twitchTrack is a 1-based track number; the multitrack output
-	// expects a 0-based mixer index for vod_track_mixer.
-	auto vod_track_mixer = (streaming->twitchVODSupported && streaming->enableTwitchVOD)
-				       ? std::optional<size_t>{osn::IAudioTrack::GetMixerIndex(streaming->twitchTrack)}
-				       : std::nullopt;
+	// Mirror SetupTwitchSoundtrackAudio's bail on an unconfigured track slot
+	// so EB silently skips VOD instead of underflowing twitchTrack=0 into a
+	// UINT32_MAX mixer index. twitchTrack is 1-based; GetMixerIndex returns 0-based.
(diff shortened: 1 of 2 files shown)


obs-studio-server/source/osn-enhanced-broadcasting-advanced-streaming.cpp
```diff
@@ -118,7 +118,19 @@ void osn::IEnhancedBroadcastingAdvancedStreaming::Start(void *data, const int64_
...
+	// encoder is actually requested when the user enables VOD Track.
+	if (streaming->enableTwitchVOD) {
+		streaming->twitchVODSupported = streaming->isTwitchVODSupported();
+	}
+
+	// streaming->twitchTrack is a 1-based track number; the multitrack output
+	// expects a 0-based mixer index for vod_track_mixer.
+	auto vod_track_mixer = (streaming->twitchVODSupported && streaming->enableTwitchVOD)
+				       ? std::optional<size_t>{osn::IAudioTrack::GetMixerIndex(streaming->twitchTrack)}
+				       : std::nullopt;
```
aleksandr-voitenko: Could we mirror the regular advanced streaming guard here before converting the selected VOD track to a mixer index? GetMixerIndex() assumes twitchTrack is a valid 1-based track number; if settings are stale/corrupt and twitchTrack is 0 or > NUM_AUDIO_TRACKS, this can underflow or pass an invalid mixer into enhanced broadcasting.

PR #1716, opened 2026-05-31
Fix UAF crash from dangling output signal handlers

summeroff:
## Problem

`osn::Output` connects the `"stop"` handler and a per-signal `OutputSignalCallback` (each carrying a heap `CallbackData` with a raw `this`) but **never disconnects them or frees the `CallbackData`**. When the underlying `obs_output_t` outlives the wrapper — delay buffer, reconnect, or a shared `SetOutput` ref — a later signal fires into freed `CallbackData`/`this` and crashes in `signal_handler_signal`.

Seen in Sentry as `EXCEPTION_ILLEGAL_INSTRUCTION` on:

```
ipc client_call_function -> ProcessPreServerCall
  -> osn::ISimpleReplayBuffer::Stop -> obs_output_stop
  -> do_output_signal -> signal_handler_signal -> [jump to freed/reused memory]
```

Heavy start/stop churn (e.g. recording repeatedly failing to write to a protected folder, with stream/recording/replay sharing encoders) widens the window.

## Fix

- Add `DisconnectSignals()`: disconnects `OnStopped` and every `OutputSignalCallback`, then frees the `CallbackData` (also fixes the pre-existing leak of those objects).
- Track allocated `CallbackData` in `m_signalCallbackData`.
- Promote the anonymous `"stop"` lambda to a named `static OnStopped` so it has a stable address that can be disconnected.
- Call `DisconnectSignals()` in `DeleteOutput()` **after** the stop-wait and **before** `obs_output_release` — `OnStopped` stays live during the 20s wait, and no connection survives the release.

All output types tear down through `DeleteOutput` (`Simple*`/`Advanced*` → `Streaming`/`Recording`/`ReplayBuffer` destructors, plus the `CreateOutput`/`SetOutput` recreate paths), so this covers every path. `signal_handler_disconnect` matches on the `(callback, data)` pair, so shared-output wrappers don't disturb each other's connections. `m_signalsReceived` is intentionally left intact so a final queued signal is still delivered to JS.

## Testing

- Existing simple/advanced replay-buffer suites pass (normal start/stop unaffected).
- clang-format clean.

### DIFF
--- a/obs-studio-server/source/osn-output.hpp
+++ b/obs-studio-server/source/osn-output.hpp
@@ -39,7 +39,6 @@ public:
 	Output(const std::vector<std::string> &signals);
 	virtual ~Output();
 
-	void ConnectSignals();
 	void CreateOutput(const std::string &type, const std::string &name);
 	void SetOutput(obs_output_t *output);
 	virtual void DeleteOutput();
@@ -72,6 +71,8 @@ private:
 	};
 
 	void InitOutput(obs_output_t *output);
+
+	void ConnectSignals();
 
 	// Disconnects every handler ConnectSignals/InitOutput wired up and frees
 	// the CallbackData. Must run while m_output is still valid.
--- a/obs-studio-server/source/osn-output.hpp
+++ b/obs-studio-server/source/osn-output.hpp
@@ -63,7 +63,21 @@ public:
 private:
 	friend void OutputSignalCallback(void *data, calldata_t *params);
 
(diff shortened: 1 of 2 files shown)


obs-studio-server/source/osn-output.hpp
aleksandr-voitenko: You can make this private.

aleksandr-voitenko (approved):

PR #1717, opened 2026-05-31
Fix volmeter use-after-free on audio thread during teardown

summeroff:
## Problem

Sentry crashes on the WASAPI real-time capture thread inside the volmeter chain (`obs_source_output_audio` → `source_signal_audio_data` → `volmeter_source_data_received`):

- **1.20.9** — `EXCEPTION_ACCESS_VIOLATION_EXEC` in `signal_levels_updated` (calling a dangling `cb.callback`)
- **1.21.2** — `EXCEPTION_ACCESS_VIOLATION_READ` in the peak path (`get_sample_peak`)

Both are a **use-after-free of the `obs_volmeter`**.

## Root cause

The volmeter holds only a **weak** reference to its attached source. When the source is concurrently going to refcount 0, `obs_volmeter_detach_source`'s weak→strong upgrade (`obs_weak_source_get_source`) returns `NULL`, so it **skips** `obs_source_remove_audio_capture_callback`. The volmeter is then freed, leaving a dangling `{volmeter_source_data_received, freed-volmeter}` entry in the source's `audio_cb_list`. The RT capture thread (not stopped until the deferred destroy phase) can call into the freed volmeter before `obs_source_destroy` frees that list.

Triggered by `Volmeter::Destroy`/`Detach` (IPC thread) racing source destruction — e.g. device removal / scene teardown. The crash logs show matching device churn (default-device change, HD60 S, Voicemod).

## Fix

Serialize the lifetime on the osn side: hold a **strong** source reference (via `findAndRef`) for the duration of the attachment, released only **after** detach. The source can no longer reach refcount 0 while attached, so the weak-ref upgrade inside `obs_volmeter_detach_source` always succeeds and the audio capture callback is removed cleanly before the volmeter is freed.

No libobs changes; blast radius stays in `obs-studio-server`.

### Tradeoff
Holding a strong ref defers a source's destruction until its volmeter is detached/destroyed. Streamlabs pairs volmeter and source teardown per audio source, so this is bounded.

### DIFF
--- a/obs-studio-server/source/osn-volmeter.cpp
+++ b/obs-studio-server/source/osn-volmeter.cpp
@@ -116,7 +116,7 @@ void osn::Volmeter::Attach(void *data, const int64_t id, const std::vector<ipc::
 		PRETTY_ERROR_RETURN(ErrorCode::InvalidReference, "Invalid Meter reference.");
 	}
 
-	auto source = osn::Source::Manager::GetInstance().find(uid_source);
+	OBSSourceAutoRelease source = osn::Source::Manager::GetInstance().findAndRef(uid_source);
 	if (!source) {
 		PRETTY_ERROR_RETURN(ErrorCode::InvalidReference, "Invalid Source reference.");
 	}
@@ -126,6 +126,7 @@ void osn::Volmeter::Attach(void *data, const int64_t id, const std::vector<ipc::
 	}
 
 	meter->uid_source = uid_source;
+	meter->source_ref = std::move(source);
 
 	rval.push_back(ipc::value((uint64_t)ErrorCode::Ok));
 	AUTO_DEBUG;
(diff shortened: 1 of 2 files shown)


obs-studio-server/source/osn-volmeter.cpp
```diff
@@ -143,6 +144,7 @@ void osn::Volmeter::Detach(void *data, const int64_t id, const std::vector<ipc::
 
 	meter->uid_source = INVALID_ID;
 	obs_volmeter_detach_source(meter->self);
```
aleksandr-voitenko, line 156: This still calls `obs_volmeter_detach_source()` while holding OSN's global `mtx`. With the new strong `source_ref`, libobs detach can now reliably reach `obs_source_remove_audio_capture_callback()`, which takes `source->audio_cb_mutex`. 

The audio thread can hold that mutex while dispatching `volmeter_source_data_received -> signal_levels_updated -> Volmeter::OBSCallback`, where OSN tries to take `mtx`. That gives a deadlock opportunity: detach thread holds `mtx` waiting for `audio_cb_mutex`, audio thread holds `audio_cb_mutex` waiting for `mtx`.

Can we split the lock scope here? Grab the `shared_ptr<Volmeter>` under `mtx`, clear OSN state as needed, release `mtx`, then call `obs_volmeter_detach_source()`, and only release `source_ref` after detach completes.

Issue #1493, opened 2024-12-04
Problem with audio when using scene


mentoRisen:
Hello,
we recently updated to new version, and that broke audio in our application. After excesive testing I manage to find out, that if I add source directly to output audio works. If I add source to scene as sceneItem, and then I add scene to output, audio is not working.
To simulate issue I have run Start advanced recording - Custom encoders in test_osn_advanced_recording.ts with following sources:
`
        let ffmpegSource = osn.InputFactory.create('ffmpeg_source', 'input02', { device_id: "ffmpeg01" });
        ffmpegSource.monitoringType = 0;
        ffmpegSource.muted = false;
        ffmpegSource.audioMixers = 1 | (1 << 2-1);

        ffmpegSource.update( Object.assign(ffmpegSource.settings, {
            local_file: "C:/reactoo/media/ForBiggerFun.mp4",
            looping: true,
            volume: 100
        }) );
        ffmpegSource.save();

        const scene = osn.SceneFactory.create('testscene');
        const sceneItem = scene.add(ffmpegSource);
        sceneItem.selected = true; 
        osn.Global.setOutputSource(1, scene); // audio does not work, mp4 is silent
        // osn.Global.setOutputSource(1, ffmpegSource); // if i comment scene and uncomment this, audio is present in mp4
`

Best Regards
Lukas Zemcak

PR #1724, opened 2026-06-04
tests: regression for scene-wrapped source audio (#1493)

summeroff:
## What

Adds an integration test reproducing #1493: an `ffmpeg_source` wrapped in a scene (with `sceneItem.video` **deliberately left unset**, so the scene item's canvas stays `NULL`) and set as the output source must still record **non-silent** audio.

Before the libobs fix this case recorded silence; with it, the source's audio is mixed through the scene as expected.

## Why it caught the bug

The existing dual-output tests pass only because they assign `sceneItem.video` (which calls `obs_sceneitem_set_canvas`). The plain `scene.add()` path — and the reporter's repro — never set it, leaving the item canvas `NULL`. This test exercises exactly that path.

## How it asserts

New `tests/osn-tests/util/media_probe.ts` reads `mean_volume` via ffmpeg's `volumedetect` filter (ffmpeg resolved from `FFMPEG_PATH` or `PATH`). Digital silence reports ~ `-91 dB`, real audio sits well above `-40 dB`, so the test asserts `mean_volume > -80 dB`. Helper logic validated locally against generated silent (`-91 dB` → fails) vs tone (`-21 dB` → passes) clips.

## Dependency

Requires the libobs fix — streamlabs/obs-studio#738 — to be present in the consumed libobs build. **This PR will not go green until that fix is rolled into the osn libobs version.**

## Notes

- Type-checks clean (`tsc --noEmit`, project options).
- Adds a new test-time dependency on an `ffmpeg` binary; OBS bundles one, or set `FFMPEG_PATH`.


summeroff:
Windows tests were failing on `spawnSync ffmpeg ENOENT` — the regression wasn't reproducing, ffmpeg just isn't on the runner. Three attempts, identical failure, so deterministic rather than flaky. macOS was unaffected since the test skips on Darwin.

Fixed by pointing `FFMPEG_PATH` at the copy that is already present: libobs ships `ffmpeg.exe`, and `obs-studio-server/CMakeLists.txt:558` installs all of `bin/64bit` next to the osn binaries. No install step, no new dependency.

Verified against the local osn package before pushing — bundled build is n7.1.1, it has the `volumedetect` filter, and the exact command `media_probe.ts` runs reports `mean_volume: -34.1 dB` for the test's own `bigbuckbunny.mp4`, comfortably above the `-80` the test asserts.

Also merged current `staging` in. Worth noting this branch was 17 commits behind and pinned to `31.1.2sl20`; it now builds against `32.1.1sl2`, so this run is the first time the test exercises OBS 32 — which is what will actually tell us whether the \#1493 fix is present there.

PR #1743, opened 2026-07-10
[osn-filter] Remove obsolete/deprecated/disabled filters

sandboxcoder:
### Description
* use logic similar to `OBSBasicFilters::CreateAddFilterPopupMenu` to remove obsolete filters
* add test validation to protect against regressions by verifying old filters no longer returned

### Motivation and Context
Update OSN to discard obsolete filters.

### How Has This Been Tested?
* Open Desktop and verified v2 of `mask_filter_v2`, `color_filter_v2`, etc are used (there is a [pull request](https://github.com/streamlabs/desktop/pull/6022) that will depend on this one).

### Types of changes

### Checklist:


sandboxcoder:
Decided to not do this approach. The issue is what if we update to OBS-33 or whatever in the future. A filter is deprecated on backend but our frontend code does not know? Well then we run into a weird edge case where user is unable to use that filter now unless we explicitly add it to the allowed list. 

For now, it just seems best to leave this alone and instead, manually update the FE allowlist when a new filter is added?

